PR #72, opened 2020-06-19
added a person wearing a hijab on the launch screen

amazingphilippe:
Dont merge if its too late. This screen will be part of the Figma demo anyways


timarney:
We can merge - the build and push doesn't happen as part of this repo.  It's a manual process at this point via xCode.

PR #101, opened 2020-06-22
Revert changes to the BackendService that are breaking our random id submission

smcmurtry:
These changes were introduced in https://github.com/cds-snc/covid-shield-mobile/commit/8416cf00e93d71b3e1e037dc7812c20f7679cebd

Not sure why they are causing our random id submission flow to fail. But probably best to remove while we figure it out.


smcmurtry:
Closing since the random id submission flow doesn't seem to be working after all.

PR #118, opened 2020-06-23
Update exposure screens

timarney:
Note - doesn't go right top the edge but we can re-visit later.

Screen Shot 2020-06-23 at 11 45 14 AM


smcmurtry:
Screen Shot 2020-06-23 at 9 53 32 AM

PR #122, opened 2020-06-23
How it works screens

nickmly:


src/screens/tutorial/TutorialContent.tsx
```diff
@@ -1,13 +1,13 @@
-import React, {useRef, useEffect} from 'react';
-import {StyleSheet, ScrollView} from 'react-native';
-import {Box, Text} from 'components';
-import {useI18n} from '@shopify/react-i18n';
+import React, { useRef, useEffect } from 'react';
+import { StyleSheet, ScrollView } from 'react-native';
+import { Box, Text } from 'components';
+import { useI18n } from '@shopify/react-i18n';
```
smcmurtry, line 4: If you are using VS code, try installing the "Prettier - Code formatter" extension. That will keep the spacing consistent with the existing codebase so we don't go back and forth on this stuff.

nickmly: Installed Prettier and used newlines, thanks for the heads up!

PR #128, opened 2020-06-23
Make onboarding headers consistent

smcmurtry:


timarney:
First slide is still lower

PR #141, opened 2020-06-24
Upstream 144

smcmurtry:
Only issue is that I'm noticing a new flicker when the bottom sheet is opened.


smcmurtry:
And we have another upstream pr to fix the flicker! I have included it here and everything looks okay.

PR #157, opened 2020-06-25
Upstream 153

timarney:


src/screens/home/HomeScreen.tsx
```diff
@@ -98,6 +98,7 @@ const Content = ({setBackgroundColor}: ContentProps) => {
       switch (systemStatus) {
         case SystemStatus.Disabled:
         case SystemStatus.Restricted:
+        case SystemStatus.Unknown:
```
smcmurtry, line 101: makes sense

timarney:
@smcmurtry let's get this one if if you agree.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -93,7 +93,7 @@ export class ExposureNotificationService {
     try {
       await this.exposureNotification.start();
     } catch (_) {
-      // Noop because Exposure Notification framework is unavailable on device
+      this.systemStatus.set(SystemStatus.Unknown);
```
smcmurtry, line 96: Going to do some local testing of this before merging

PR #160, opened 2020-06-25
Add english instructions for fixing gem error

CalvinRodo:
I ran into an issue while following the install guides with my default CDS Macbook setup. 

I documented the steps I took to fix it, hopefully this can save someone else from having to research a fix for the same issue in the future.


timarney:
Thanks @CalvinRodo

PR #248, opened 2020-06-28
Added countdown text

smcmurtry:
Screen Shot 2020-06-27 at 9 07 19 PM
Screen Shot 2020-06-27 at 9 07 04 PM

@EliseKa When there is one day left should this be "Il reste 1 jour sur 14!"?


EliseKa:
@smcmurtry 
yes you're right, when there is one day left it should be "Il reste 1 jour sur 14!

PR #284, opened 2020-06-29
Onboarding margins

nickmly:
Smaller margins on all onboarding views


nickmly:
@smcmurtry I removed all margins from the views. It looks good on an iPhone 8 and iPhone 11

PR #288, opened 2020-06-29
Diagnosis button update

its-anirudh:


its-anirudh:
> It looks like this pr contains a bunch of privacy policy updates? Could you move those to your privacy policy pr #279 instead?
> 
> Also, the functionality to include a chevron was already part of the button component - you just need to set the `internalLink` prop on the button. Please use that instead.

When I used it, the internalLink icon's color is the same color as the button in the design, so that approach will not work here. Hence the need for a custom icon.

smcmurtry:
> When I used it, the internalLink icon's color is the same color as the button in the design, so that approach will not work here. Hence the need for a custom icon.
Okay good point @its-anirudh! You can disregard that part of my review.

PR #335, opened 2020-06-30
French updated

its-anirudh:


timarney:
Using https://github.com/cds-snc/covid-shield-mobile/pull/341

PR #354, opened 2020-07-02
Upstream 180

timarney:
https://github.com/CovidShield/mobile/pull/180


timarney:
Agreed on the  try/catch/ etc...

@jeberhardt mentioned as well.

Issue #337, opened 2020-06-30
Not able to activate the "Next" button from step 5 to step 6 with Talkback


sarasuri:
Scenario to reproduce

Samsung Galaxy J3 Prime
Android version 7.0 and 8.0

Step 1 to 6 screens after splash screen

While Talkback is on, swiping right to move to next items from Step 1 screen, Talkback moves to each screen and continue reading. On the step 6 screen, pressing “Next” button, takes users to step 2 screen, swipe again to move to each item, Talkback reads steps 2 to 6 again. The same behaviour with each subsequent step.

There are other issues as well.

1.	When Talkback moves to each screen from Step 1, the step number is not updated at the bottom to reflect what screen is being read.

2.	Step 4 screen “How it works”, it only reads from second paragraph (Every day, it checks …) 
Step 5 screen, it reads from first paragraph missing the heading of the screen
Step 6 screen, it appears focus is on Nova Scotia and only reads from that point

3.	After step 4 screen, pressed “Next” button on the step 6 screen to go to step 5 screen.
When Talkback on step 5 screen, swipe to move to each element and Talkback moves to step 6. On the step 6 screen, double tapping on “Next” button doesn’t activate it and users are stuck on the screen. They can go back to previous screen using the “Back” button but not able to move forward using the “Next” button. 

Expected results
Content should be read in a proper order. Users should be able to activate the "Next" button.

Screenshots
![step4](https://user-images.githubusercontent.com/12256067/86174773-40c82f80-baf0-11ea-9413-28b46cc9a13d.gif)
![step5](https://user-images.githubusercontent.com/12256067/86174777-4291f300-baf0-11ea-911f-4c45f119fcb3.gif)

qanitah:
Confirming the same behaviour on Android 10 (OnePlus 7).

JuliannaR:
This is fixed in V15.

Issue #317, opened 2020-06-30
Content is not read in correct order when moving from screen 4 to 5 with Talkback


sarasuri:
Scenario to reproduce

Samsung Galaxy J3 Prime
Android version 7.0

Step 1 to 6 screens after splash screen

While Talkback is on, swiping right to move to next element on the screens from Step 1, Talkback moves to each screen and continue reading. When it moves to step 5 screen “The app will now ask your permission”, it announces “The app uses Bluetooth to exchange random …” which is from the previous screen. When continue swiping the focus goes to “Français” button and read from there.

Expected result
Talkback should read the heading of the step 5 screen like it was reading for other screens.

WCAG
https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html

Screenshots
![screen5b](https://user-images.githubusercontent.com/12256067/86148713-faaba580-bac8-11ea-8853-816644112dd1.gif)
![screen5a](https://user-images.githubusercontent.com/12256067/86149106-81608280-bac9-11ea-9b99-528f55f11c33.gif)

qanitah:
The same behaviour for Samsung Android version 8.

nickmly:
This seems to be an issue with the `BulletPointX` component on the `Anonymous` view in Onboarding. Looking into it

JuliannaR:
The good news, I can access all the content in VoiceOver in this update. There is an issue though, the focus when you go from screen to screen is on the button and not the heading, can we fix that @nickmly

JuliannaR:
This is fixed in your latest PR around onboarding flows. @smcmurtry the screen has changed a lot but the reading order was addressed re: page nav. Can you confirm and we can close?

smcmurtry:
Yes my pr #398 fixed this issue

Issue #304, opened 2020-06-30
Steps are not announced by Talkback (Android)


sarasuri:
Scenario to reproduce

Samsung Galaxy J3 Prime
Android version 7.0

Step 1 to 6 screens after splash screen

While Talkback on, using “Read from top” from “Global Context menu” Talkback starts reading from the top of the screen and moves to each step automatically and continue reading. However, it doesn’t announce what step it is reading. At the end of the step 6, it announces “Step 1 of 6 and “Next” button. Further, the same behaviour when tried swiping through each element. 

Expected result
Talkback should announce each step.

Screenshots
![screen6](https://user-images.githubusercontent.com/12256067/86131653-ead59680-bab3-11ea-92c6-3a6985df49ae.gif)

JuliannaR:
Fixed in IOS V16.

PR #367, opened 2020-07-02
Onboarding accessibility

nickmly:
Fixes the following issues: #304 #317 #337 

The onboarding screen is now accessible using a screen reader from end to end. Tested on Android using Talkback.

Changes:
- Switched off of `Carousel` component to `FlatList` (had many issues with performance and accessibility with the Carousel)
- Moved the pagination and next button to the `renderItem` callback so that screen readers will read them out
- Fixed screen readers getting stuck on steps that had `BulletPoint` components in them

### DIFF
--- a/src/screens/onboarding/Onboarding.tsx
+++ b/src/screens/onboarding/Onboarding.tsx
@@ -37,16 +37,43 @@ export const OnboardingScreen = () => {
   const startExposureNotificationService = useStartExposureNotificationService();
   const maxWidth = useMaxContentWidth();
 
+  const isStart = currentIndex === 0;
+  const isEnd = currentIndex === contentData.length - 1;
+
+  let endText = 'EndSkip';
+  let endBtnStyle = '';
+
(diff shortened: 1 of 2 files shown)


smcmurtry:
I just had a thought - this carousel stuff is complicated and I don't think it's actually doing anything for us at this point. 

What if we just use pages here? Every onboarding step becomes its own page inside the `<OnboardingStack.Navigator>`. We are not using modal style navigation for the onboarding navigator, we are using regular navigation with a fade transition: https://github.com/cds-snc/covid-shield-mobile/blob/master/src/navigation/MainNavigator.tsx#L73. So I don't think those transitions would be much different from the carousel.

timarney:
@smcmurtry @nickmly - noting the solution will need to work for onboarding and how it works.

JuliannaR:
@smcmurtry regular nav could work. There are issues on most pages that require scrolling at the moment.

smcmurtry:
Closing in favour of #385, #398

PR #369, opened 2020-07-02
Alt text gh330

its-anirudh:


JuliannaR:
@its-anirudh These look good. Can you update to include the how it works flow. ALT text is also in figma. 

Image from iOS (7)


Image from iOS (6)

JuliannaR:
We still need to add the ones from the how it works flow, they dont appear in the changed files?

its-anirudh:
> We still need to add the ones from the how it works flow, they dont appear in the changed files?

I will add them; did not see them in scope in the GH ticket

PR #383, opened 2020-07-03
Fix launch_screen temporarily

henrytao-me:
Temporarily fix for splash screen crash by generating different size for existing launch_screen image 

Part of https://github.com/cds-snc/covid-shield-mobile/issues/382


henrytao-me:
I will do a follow up PR to make splash screen responsive and use svgs but this is good to unblock for now.

PR #385, opened 2020-07-03
Removed duplicate SafeArea in RegionPicker

nickmly:


smcmurtry:
I tested your PR - using voiceover I am able to scroll on the region page in current master and in your PR… So I guess this is not an issue. I must have misunderstood the issue @JuliannaR  was raising here.

PR #387, opened 2020-07-03
Collapse BottomSheet after submission flow is finished

henrytao-me:
This PR collapses BottomSheet after submission flow is finished by: 
- Expose BottomSheet using forwardRef https://reactjs.org/docs/hooks-reference.html#useimperativehandle
- Compare previous and current ExposureStatus to determine if the BottomSheet needs to be collapsed. This is not a generic solution but it is simple and prevent nested access to children views. 

Resolves https://github.com/cds-snc/covid-shield-mobile/issues/76

### DIFF
--- a/src/components/BottomSheet/BottomSheet.tsx
+++ b/src/components/BottomSheet/BottomSheet.tsx
@@ -23,31 +23,26 @@ export interface BottomSheetBahavior {
   collapse(): void;
 }
 
-const BottomSheet = (
+const BottomSheetInternal = (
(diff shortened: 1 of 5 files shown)


henrytao-me:
It's separate. This is purely for closing BottomSheet automatically after submission flow. There is another issue for what you mention https://github.com/cds-snc/covid-shield-mobile/issues/292

JuliannaR:
@henrytao-me - thanks for clarifying. I linked them earlier today.

PR #392, opened 2020-07-03
Fix close button in Tutorial screen and improve performance

henrytao-me:
This PR fixes close button in Tutorial screen and improve performance at the same time.

Resolves https://github.com/cds-snc/covid-shield-mobile/issues/360


nickmly:
Looks good on a smaller screen but I am seeing extra padding at the top on a larger screen:

### Nexus 4:
Screenshot_1593805696

### Pixel 3 XL:
Screenshot_1593805695

timarney:
Going to merge - we'll fix up the spacing a bit in another PR

henrytao-me:
The reason for extra padding on the top is because we do center vertical alignment. I don't think we need to fix. 🤔

PR #394, opened 2020-07-03
How it works a11y fixes

nickmly:
- Replaced `Carousel` with `FlatList` component for performance
- Fixed images flickering between screens
- All content is read through a screen reader properly
- Steps now increment when a screen reader is reading through each screen
- Body text is separated into different `Text` components

### DIFF
--- a/src/screens/tutorial/views/ItemView.tsx
+++ b/src/screens/tutorial/views/ItemView.tsx
@@ -12,6 +12,7 @@ interface ItemViewProps {
 
 export const ItemView = ({item, image}: ItemViewProps) => {
   const [i18n] = useI18n();
+  const bodyText = i18n.translate(`Tutorial.${item}`).split(/\n\n/g);
   return (
     <Box>
       {image && (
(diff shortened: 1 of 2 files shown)


nickmly:
@smcmurtry Could you verify if this works on iOS? I was able to get it working with TalkBack and had no issues going through the flow.

src/screens/tutorial/views/ItemView.tsx
```diff
@@ -1,27 +1,37 @@
...
 
 interface ItemViewProps {
   item: TutorialKey;
-  image: ImageSourcePropType;
+  image?: ImageSourcePropType;
 }
 
 export const ItemView = ({item, image}: ItemViewProps) => {
   const [i18n] = useI18n();
+  const bodyText = i18n.translate(`Tutorial.${item}`).split(/\n\n/g);
```
smcmurtry, line 15: Clever!

smcmurtry:
@nickmly as far as I'm aware we don't really have a problem with the carousel here. But I do want to get in the text splitting fix. Could you open another PR for that and we'll close this one.

PR #406, opened 2020-07-04
Split body text in tutorial by newlines

nickmly:
Body text in the tutorial screens should now be read as separate text elements by screen readers


src/screens/tutorial/views/ItemView.tsx
```diff
@@ -12,6 +12,7 @@ interface ItemViewProps {
 
 export const ItemView = ({item, image}: ItemViewProps) => {
   const [i18n] = useI18n();
+  const bodyText = i18n.translate(`Tutorial.${item}`).split(/\n\n/g);
```
timarney, line 15: Nice :)

PR #415, opened 2020-07-04
Larger touch target for back button in Toolbar on Android

### LINKED ISSUE #413: Android - Increase back button size / touch target - How it works
Screen Shot 2020-07-04 at 11 11 54 AM

nickmly:
Screenshot_1593883104
Fixes #413

### DIFF
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -30,7 +30,7 @@ export const Toolbar = ({title, navText, navIcon, navLabel, onIconClicked}: Tool
   if (Platform.OS === 'android') {
     return (
       <Box flex={1} flexDirection="row" alignItems="center" justifyContent="flex-start" padding="none" maxHeight={56}>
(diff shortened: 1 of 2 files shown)


timarney:
@nickmly can you drop a screenshot into the comment above?

nickmly:
@timarney Updated the comment. The button is the same size but the target is larger. I can also update the button size if needed.

emilykuret:
Thinking we need to update the button size. On my Pixel XL it's quite small. Google Suggested min. 48 pixels if I remember correctly.

nickmly:
Updated to be 48px in size: 
Screenshot_1593884733

timarney:
@emilykuret ☝️ see button above and comment.

emilykuret:
There seems to be a missing pixel in the arrow - but the size looks good. thank you!

timarney:
Okay going to merge this as is as it covers the original ask.

Additional issues should be open pending direction on 

1) Should it be text?
https://github.com/cds-snc/covid-shield-mobile/issues/250

2) If not - fix the back arrow icon so it can be bigger.

PR #422, opened 2020-07-05
Sentry

timarney:


package.json
```diff
@@ -24,9 +24,7 @@
     "@react-navigation/drawer": "^5.7.2",
     "@react-navigation/native": "^5.3.0",
     "@react-navigation/stack": "^5.3.2",
-    "@sentry/integrations": "^5.19.0",
     "@sentry/react-native": "^1.5.0",
-    "@sentry/wizard": "^1.1.4",
```
jeberhardt, line 29: While **@sentry/wizard** can be left out, **@sentry/integrations** needs to be in, to allow for the **CaptureConsole()** to work.

jeberhardt: Since it's still being imported in `src/index.ts`.

PR #426, opened 2020-07-05
Header test mode

### LINKED ISSUE #215: Unable to use 2-fingers-up gesture on some screens to read full page from top when VoiceOver on
While VoiceOver is on and 2 fingers up gesture is used to read the page from top, it doesn't work for most screens. When 2 fingers are flicked up on a screen, it should read all content on the page, starting from top.

It seems to work on screens that don't have the "COVID Alert" image on top of the page. 

Please see screenshots of one screen where it works and one where it doesn't.
iOS 13.5.1 on iPhone 7

covidalert5
covidalert4

smcmurtry:
For release, we will set `TEST_MODE = 'false'` in the .env file, and this pr will "fix" part of  #215.

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -2,6 +2,7 @@ import React, {useCallback} from 'react';
 import {TouchableWithoutFeedback} from 'react-native';
 import {useNavigation, DrawerActions} from '@react-navigation/native';
 import {useI18n} from '@shopify/react-i18n';
+import {TEST_MODE} from 'env';
 
 import {Box} from './Box';
 import {Icon} from './Icon';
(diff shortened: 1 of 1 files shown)


timarney:
I think this is a part solve and do think it should be merged.

It doesn't solve:

The cancel button - same issue here.

85895623-08a9af80-b7c5-11ea-9f75-83a6f7889dd8

@JuliannaR per #215 the issue is for "TEST" mode the logo is a button which it won't be for release - Steve correct that here.

If the app is in Test mode - the issue will still exist.

I don't want to spend much time on a fix that will end up going away ... Steve we'll still need to address the cancel button.

@JuliannaR @smcmurtry let's chat about this (today hopefully) to see what we can do to get that issue closed.

timarney:
Steve will fix the other part in a separate PR

Issue #257, opened 2020-06-28
Menu: Android does not display notifications are off when COVID Alert is OFF


mmiron-HC:
Android v9 and 8
samsung Galaxy S8+ and OnePlus 7
- COVID-19 Exposure logging = off
- Bluetooth : Off
- Notifications : Off

1. Install the app for the first time 
2. Navigate through the Onboarding screens until the Enable COVID-19 exposure logging and notifications is displayed.
3. Click Don't Enable
4. Verify that the menu displays the following messages:  
EN: COVID Alert is OFF and Notifications are OFF 
FR:  Alerte COVID est désactivé and Notifications désactivées

iPhone displays both messages
Android does not
Notification message should be displayed like iOS.

![Android_Menu_no notifications msg](https://user-images.githubusercontent.com/65728700/85949026-39532b80-b922-11ea-9c5e-ae06a79c8d1f.jpg)

smcmurtry:
This should now be fixed. The logic that decides when to display the warning in the collapsed bottom sheet has changed. There used to be cases where it would show up in the expanded sheet, but not in the collapsed sheet. 

There is a related issue where this content isn't updating as frequently as it needs to if you make a change via the settings screen. I will file an issue for that separately.

JuliannaR:
@mmiron-HC can you verify that this works for you now based on @smcmurtry comment above? I sent you the link for a playstore download of the app with our meeting invite.

timarney:
We need to push to the Play store again first :) 

Plan to see this change first thing in the A.M

mmiron-HC:
both messages now appear tested using OnePlus 3
![Screenshot_20200709-064537](https://user-images.githubusercontent.com/65728700/87030736-09bce280-c1b0-11ea-9397-9d3ff250e712.jpg)

PR #429, opened 2020-07-05
Change initial notifications status on Android.

jeberhardt:


timarney:
App Covid Alert 

Status says: **OFF**



<hr>

App Covid Alert 

Status says: **OFF**



<hr>

Settings App Covid Alert 

Status says:  **ON**



Possibly different on the Play Store 

@henrytao-me thoughts on this PR?

jeberhardt:
If I'm interpreting this issue correctly, here's a screenshot of the "Notifications" status that isn't showing up on Android. This is a screenshot taken with this PR applied. 

Screenshot_20200705-184118

timarney:
So I think it's Push Notifications and EN notifications are off.

Either way we should be a side by side comparison. 

From the issue:

Verify that the menu displays the following messages:
EN: COVID Alert is OFF and Notifications are OFF
FR: Alerte COVID est désactivé and Notifications désactivées

iPhone displays both messages
Android does not

We could side load the app - "fresh install" and do the same actions to compare the results.

timarney:
This other issue is part of this entire flow as well.

https://github.com/cds-snc/covid-shield-mobile/issues/271

timarney:
Okay - I think @smcmurtry did some code around this as well that never got into the app.

Given whatever we do will effect both devices - 

A) @smcmurtry 

Seems like this would be a better initial state

```javascript
const [status, setStatus] = useState<NotificationPermissionStatusContextProps['status']>('blocked');
```

If yes let's merged this and see if the the testers do a side by side comparison.  From a fresh install - on the current Play Store + TestFlight App.

I would like to see if taking the same actions - on each device ends up giving the same results.  Paying particular attention to the order in which things are done.

Reason if we're changing the code here it won't just be for Android.

We also need to fix this one which relies on the status as well.

https://github.com/cds-snc/covid-shield-mobile/issues/271

smcmurtry:
@timarney and I had a chat about this and I think the issue #257 that this is addressing was already fixed in an unrelated PR. Although there are some other issues with the message not updating after a change on the app settings screen that still need to be addressed.

PR #431, opened 2020-07-05
How it works - use same nav "cancel" for Android and iOS

timarney:
**Note:** Text can be updated in a separate PR

This makes the PR makes the nav consistent.

**iOS**
Screen Shot 2020-07-05 at 10 29 49 AM

<hr>

**Android**


JuliannaR:
Yes, this will make IOS and Android match. @amazingphilippe might have thoughts for later, but this will be consistent.

PR #447, opened 2020-07-05
WIP - init Fastlane debug build Android

timarney:


timarney:
@henrytao-me 

Can adjust this further but dropping this as a start.

timarney:
@dsamojlenko this is down on the priority list but we should get this in at some point.

At the moment flipping this line between Debug and Release
https://github.com/cds-snc/covid-shield-mobile/blob/d423aa4b26780068d36ece61c71feca01a2b272d/fastlane/Fastfile#L84

Issue #276, opened 2020-06-29
No focus on the 8-digit code form field (Android)


sarasuri:
Scenario to reproduce
Samsung Galaxy J3 Prime
Android version 7.0

When users touch down on the form field to enter an 8-digit code on the “Enter your 8-digit number” screen there is no focus on the field.

Expected result
The form field should given focus for users to see where they are.

"WCAG
https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html

Screenshots

![focus](https://user-images.githubusercontent.com/12256067/86004355-2eab9b80-b9e1-11ea-9dff-3b641a3aeb88.gif)

emilykuret:
I think this is on iOS as well

smcmurtry:
So the issue here is that there is no styling to indicate the field is in focus after it is tapped. We would need some designs for what the focus style should look like before we do anything here.

jeberhardt:
@smcmurtry, while I agree about the styling, I think this issue all by itself is valid, and can be closed as is. The styling is a separate issue that applies whether the user taps in the field, or it is automatically selected.

timarney:
The form field should given focus for users to see where they are.

> "...see where they are"

a) Per @smcmurtry  
   We would need some designs for what the focus style should look like before we do anything here.

<img width="200" alt="Screen Shot 2020-07-05 at 8 45 26 AM" src="https://user-images.githubusercontent.com/62242/86532980-ea664280-be9b-11ea-861d-54705ae711aa.png">

☝️ Steve you flagged issues here before as the input focus isn't really multiple inputs it's a single input styled to look like multiple inputs - this should be discussed again.

b) I like what @jeberhardt did as far as Auto-focus (pops the Keyboard and gives focus to the input) but as @smcmurtry 

notes: 

> ...I think we should probably auto focus the headers on both this and the next page, so that focus starts from the top of the page when the user navigates to the page.

Does the non screen reader view in this case make the screen reader experience bad cc: @JuliannaR

JuliannaR:
I agree with the auto focus and giving focus to the input. Because there is some visual trickery that happens in the physical design as its actually all one input, can we focus on it like its just 1 element? @smcmurtry

timarney:
Talked to @JuliannaR 

1) Give focus to the header
2) Update the styling "for all types of users" on the input - perhaps just dim the underline for today to indicate that yes that part of the input has focus (no really but visually it does)
[...]

PR #448, opened 2020-07-05
Make the code input styling easier to see

smcmurtry:
"Closes" #276
This styling was already there but it was invisible because the lines were too narrow and the blue being used was too dark. So I made it CDS yellow:

code_input

It is not perfect because the yellow styling is applied before the input is actually in focus. We plan to revisit this later.


timarney:
I'll review this one.

PR #460, opened 2020-07-06
Splash screen alt text

nickmly:
Currently only works on Android.


timarney:
@nickmly 

We'll revisit this after the v1 launch - 

We need to fix this first
https://github.com/cds-snc/covid-shield-mobile/issues/382

PR #468, opened 2020-07-06
Fix accessibility in BottomSheet

henrytao-me:
This PR fixes (hopefully all) accessibility issues related to BottomSheet and content underneath. 

To prevent content underneath being read, we need to use `accessibilityElementsHidden` (for iOS) or `importantForAccessibility` (for Android) when BottomSheet is expanded.  

Resolves https://github.com/cds-snc/covid-shield-mobile/issues/432


henrytao-me:
Tested on both Android and iOS. I merge this for now. We can do follow up / revert PR if needed.

PR #481, opened 2020-07-07
Changed 'Cancel' on pages to 'Close' in toolbar

### LINKED ISSUE #479: On the province and language change pages the link to close / confirm says cancel
The only option on pop-ups is to “Cancel” — when you change your province for instance, perhaps Cancel implies the change doesn’t apply? In Figma, it says “Close”
In these cases it needs to say close so that people know they change has been accepted. 

Image from iOS (13)

nickmly:
Fixes #479


src/locale/translations/fr.json
```diff
@@ -162,7 +162,7 @@
     "InformationTitle": "À propos d’Alerte COVID"
   },
   "LanguageSelect": {
-    "Close": "Annuler",
+    "Close": "Retour",
```
smcmurtry, line 165: Should this be "Fermer" to be consistent with english?

PR #493, opened 2020-07-07
Removed onboarding transition animation

smcmurtry:


timarney:
Approving this knowing 

> We're reverting back to the carousel for the on-boarding screens (@henrytao-me will have a PR later for this)

https://github.com/cds-snc/covid-shield-mobile/issues/490

PR #495, opened 2020-07-07
Back color adjust

jelen-HC:
Adjusted the Back Button Color to the correct blue color for the Onboarding and How it works sections.


timarney:
Ref: https://github.com/cds-snc/covid-shield-mobile/issues/478

PR #496, opened 2020-07-07
Fix accessibility in carousel

henrytao-me:
This PR fixes accessibility in Carousel, including: 
* Scope accessibility down to page instead of screen.
* Run accessibility for each sentences.
* Auto focus to first element after each Next which I think it's the Image instead of Cancel button.

Ref https://github.com/cds-snc/covid-shield-mobile/issues/490

### DIFF
--- a/src/screens/tutorial/views/ItemView.tsx
+++ b/src/screens/tutorial/views/ItemView.tsx
@@ -1,4 +1,4 @@
-import React, {useRef, useEffect, useLayoutEffect} from 'react';
+import React, {useRef, useLayoutEffect} from 'react';
 import {StyleSheet, Image, ImageSourcePropType, AccessibilityInfo, findNodeHandle} from 'react-native';
 import {Text} from 'components';
 import {useI18n} from '@shopify/react-i18n';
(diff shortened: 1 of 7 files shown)


timarney:
@henrytao-me - 

@smcmurtry will review - I'll be out for a bit.

src/screens/tutorial/views/ItemView.tsx
```diff
@@ -1,21 +1,32 @@
...
-export const ItemView = ({item, image}: ItemViewProps) => {
+export const ItemView = ({item, image, isActive}: ItemViewProps) => {
   const [i18n] = useI18n();
   const bodyText = i18n.translate(`Tutorial.${item}`).split(/\n\n/g);
+  const imageRef = useRef<any>();
+
+  useLayoutEffect(() => {
+    const tag = findNodeHandle(imageRef.current);
+    if (isActive && tag) {
+      AccessibilityInfo.setAccessibilityFocus(tag);
```
smcmurtry, line 22: So that's how you do that! Awesome. We will be using this other places in the app as well.

henrytao-me: This is for auto focus. The but main fix for Carousal is `importantForAccessibility="no"` for Android and `accessibilityElementsHidden={index !== currentStep}` for iOS

Issue #504, opened 2020-07-07
2 finger swipe up to read full page not working with VoiceOver


qanitah:
With VoiceOver on, when the user does 2 finger swipe up for this page, the focus stays on the COVID Alert image on top and the page is not read as its expected to.

Screenshot below shows where the focus stays and nothing is read/announced on the page.

iOS 13.5.1 on iPhone 7

![image](https://user-images.githubusercontent.com/14280062/86838298-fefa3480-c06d-11ea-9b69-01d04ae89baa.png)

![image](https://user-images.githubusercontent.com/14280062/86838897-b8590a00-c06e-11ea-93df-dd026326f28c.png)

timarney:
This is cause because the logo is a button for the test mode menu - 

We're going to start pushing up builds without that menu on and this problem will go away.

JuliannaR:
Yes, this works if you deselect the COVID alert button. It won't be a button on release because its currently used to access the tester nav options.

PR #512, opened 2020-07-07
Fixing two finger swipe for VO

mtropper:
Issue #504 2 finger swipe up to read full page not working with VoiceOver #504


timarney:
Yes - the test demo will get removed but ... seems folks are relying on it at this point.

I'll take a look at this as soon as I can.

RE: the test menu I've floated the idea to @smcmurtry of adding a test menu button into the bottom drawer nav which which would be turned off for release.  have a button as the last item in that menu is going to be better than wrapping the logo which is causing many issues at this point.

mtropper:
The fix breaks the link button. Deleting the fix

PR #524, opened 2020-07-08
Handling the `reject` of getRandomBytes for both iOS and Android.

jeberhardt:
In both of the cases where getRandomBytes is called, if an error occurs, it will be logged so we know it happened, and thrown again and caught by the error handlers of FormView and ConsentView.

### DIFF
--- a/android/app/src/main/java/app/covidshield/module/CovidShieldModule.kt
+++ b/android/app/src/main/java/app/covidshield/module/CovidShieldModule.kt
@@ -28,9 +28,13 @@ class CovidShieldModule(context: ReactApplicationContext) : ReactContextBaseJava
     @ReactMethod
     fun getRandomBytes(size: Int, promise: Promise) {
         promise.launch(this) {
-            val bytes = SecureRandom().generateSeed(size)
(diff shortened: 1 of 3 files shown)


jeberhardt:
@obrien-j - In the rest of the ***ts** code, `catch(error)` is everywhere. I used **exception** in the Kotlin file though. ;)

android/app/src/main/java/app/covidshield/module/CovidShieldModule.kt
```diff
@@ -28,9 +28,13 @@ class CovidShieldModule(context: ReactApplicationContext) : ReactContextBaseJava
     @ReactMethod
     fun getRandomBytes(size: Int, promise: Promise) {
         promise.launch(this) {
-            val bytes = SecureRandom().generateSeed(size)
-            val base64Encoded = Base64.encodeToString(bytes, Base64.DEFAULT)
-            promise.resolve(base64Encoded)
+            try {
```
henrytao-me, line 31: There is no need to try / catch here because `.launch` extension already handle exception.

Maybe next time, please ping me to PR for review whenever we change native code as I would like to bring it over to original repo too. 🙇 

cc @jeberhardt @timarney

PR #536, opened 2020-07-08
Fastlane: Force provisioning profile for AppStore builds

dsamojlenko:
This just forces a Provisioning Profile on AppStore builds, like we do on AdHoc. To prevent auto-matching with incompatible profiles.


timarney:
@dsamojlenko ☝️

dsamojlenko:
No longer needed, closing

PR #540, opened 2020-07-08
Update NSAppTransportSecurity values.

jeberhardt:
This updates the iOS NSAppTransportSecurity values for better security between client and server.

While the ** NSAllowsLocalNetworking** key shouldn't be necessary, I have found that the simulator doesn't connect to the React-Native Metro server without that key.

### DIFF
--- a/ios/CovidShield/Info.plist
+++ b/ios/CovidShield/Info.plist
@@ -35,16 +35,10 @@
 	<true/>
 	<key>NSAppTransportSecurity</key>
 	<dict>
-		<key>NSAllowsArbitraryLoads</key>
+		<key>NSAllowsLocalNetworking</key>
 		<true/>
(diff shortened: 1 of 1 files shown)


timarney:
Will hold off on merging this until after the next builds / reviews.

timarney:
Moving this to draft for now.

timarney:
@obrien-j

If you want to take a look we can get this in today.

timarney:
Looks good on my end - 

@obrien-j please have a look when you can.

Screen Shot 2020-07-10 at 7 17 07 PM

timarney:
Going to merge and we'll adjust as needed.

obrien-j:
Thanks all. This is good. I'll double check on the need for AllowsLocalNetworking, given my and James read through of the docs, but it's great as is. Stamping for security review.

Issue #462, opened 2020-07-06
English app name and icon on French phone


katewilhelm:
When a person's language setting on their phone is set to French, the app icon on the home screen still shows the name in English. 
Also, the OS messages still use the English name when the phone is set to French. They should all have the French name

![IMG-0374](https://user-images.githubusercontent.com/57191279/86633737-09013200-bf9f-11ea-82be-d4626d8988bc.PNG)
![IMG-0377](https://user-images.githubusercontent.com/57191279/86633771-14545d80-bf9f-11ea-802a-94c129f434b6.PNG)
![IMG-0375](https://user-images.githubusercontent.com/57191279/86633787-18807b00-bf9f-11ea-80b2-eccf13861d06.PNG)
![IMG-0376](https://user-images.githubusercontent.com/57191279/86633795-1b7b6b80-bf9f-11ea-9fed-3156836852e4.PNG)

mtropper:
This isn't a development issue. The apps details haven't been entered into the Apple app store yet. There is a lot of content that is missing from the app store listing that will need to be completed before the app can go live, such as bilingual screenshots of the app and app store description etc.

mtropper:
I added the French title for the app but both English and French are missing the details of the app in both languages. I'd imagine its been planned to be entered once the app is ready to be released.

timarney:
>I added the French title for the app but both English and French are missing the details of the app in both languages. I'd imagine its been planned to be entered once the app is ready to be released.

@katewilhelm

timarney:
> This isn't a development issue. The apps details haven't been entered into the Apple app store yet. There is a lot of content that is missing from the app store listing that will need to be completed before the app can go live, such as bilingual screenshots of the app and app store description etc.

☝️ noting this

@Stephen-PM

mtropper:
@timarney @Stephen-PM I double checked the localization settings in the app, and the French  title has not been entered and the InfoPlist.string file needs to be created for the app to display name properly in French

Stephen-PM:
I added the French app name onto both App stores.

PR #541, opened 2020-07-08
English app name and icon on French phone #462

mtropper:
Issue #462 Added French Name for IOS and Android


jeberhardt:
On an iPhone SE, I changed the language to **Français (Canada)**. I installed a fresh copy of the app, and the name appeared as Alerte COVID, as expected.

PR #549, opened 2020-07-08
Add accessibilityAutoFocus to Text component

henrytao-me:
Follow up of https://github.com/cds-snc/covid-shield-mobile/pull/534

### DIFF
--- a/src/components/Text.tsx
+++ b/src/components/Text.tsx
@@ -1,18 +1,20 @@
-import React, {forwardRef} from 'react';
+import React from 'react';
 import {TextProps as RestyleTextProps, textRestyleFunctions, useRestyle} from '@shopify/restyle';
 import {Text as RNText} from 'react-native';
 import {Theme} from 'shared/theme';
+import {useAccessibilityAutoFocus} from 'shared/useAccessibilityAutoFocus';
 
 // See https://github.com/Shopify/restyle/blob/master/src/createText.ts
 export type TextProps = RestyleTextProps<Theme> &
-  Omit<React.ComponentProps<typeof RNText> & {children?: React.ReactNode}, keyof RestyleTextProps<Theme>>;
+  Omit<React.ComponentProps<typeof RNText> & {children?: React.ReactNode}, keyof RestyleTextProps<Theme>> & {
+    accessibilityAutoFocus?: boolean;
+  };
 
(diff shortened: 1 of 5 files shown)


src/components/Text.tsx
```diff
@@ -1,19 +1,21 @@
...
 
-const BaseText = (props: TextProps, ref?: React.LegacyRef<any>) => {
+export const Text = ({accessibilityAutoFocus = false, ...props}: TextProps) => {
   const styledProps = useRestyle(textRestyleFunctions, props);
-  return <RNText {...styledProps} ref={ref} />;
+  const autoFocusRef = useAccessibilityAutoFocus(accessibilityAutoFocus);
+  return <RNText accessible ref={autoFocusRef} {...styledProps} />;
 };
 
-export const Text = forwardRef(BaseText);
```
henrytao-me, line 15: There is no need for forwardRef when we use prop.

nickmly:
Works great on Android!

src/shared/useAccessibilityAutoFocus.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+const AUTO_FOCUS_DELAY = 200;
+
+/**
+ * Remember to add `accessible={true}` to target component
+ */
+export const useAccessibilityAutoFocus = (isActive = true) => {
+  const [autoFocusRef, setAutoFocusRef] = useState<any>();
+  useLayoutEffect(() => {
+    if (!autoFocusRef || !isActive) {
```
henrytao-me, line 18: First, we don't want to do this on unnecessary components.

src/shared/useAccessibilityAutoFocus.ts
```diff
@@ -0,0 +1,47 @@
...
+      return;
+    }
+    const {callable, cancelable} = createCancellableCallbackPromise(
+      () => AccessibilityInfo.isScreenReaderEnabled().then(delay(AUTO_FOCUS_DELAY)),
```
henrytao-me, line 22: Second, we only do the timeout and focus if screenReader is enabled.

henrytao-me: Adding the timeout 200ms seems to be more stable. `createCancellableCallbackPromise` will make sure if component is unmounted before timeout is finished, it won't leak.

PR #555, opened 2020-07-08
Improve detect exposure flow

henrytao-me:
Using timeout can introduce race condition. This PR change the implementation on Android by:
- Create a Promise wrapper to feed all 14 keys to the framework.
- If one of the request is returned with matched key count, the Promise will be resolved. 
- We still keep the logic of `Not persisting lastCheckPeriod` for Android only. So that we won't miss any cases.

Follow up of https://github.com/cds-snc/covid-shield-mobile/issues/453

### DIFF
--- a/android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
+++ b/android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
@@ -14,7 +14,6 @@ import app.covidshield.extensions.toInformation
 import app.covidshield.extensions.toSummary
 import app.covidshield.extensions.toWritableArray
 import app.covidshield.extensions.toWritableMap
-import app.covidshield.extensions.withDelay
 import app.covidshield.models.Configuration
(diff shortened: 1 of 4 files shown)


android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
```diff
@@ -105,23 +102,15 @@ class ExposureNotificationModule(context: ReactApplicationContext) : ReactContex
 
             exposureNotificationClient.provideDiagnosisKeys(files, exposureConfiguration, token).await()
 
+            // Clean up files after feeding to the framework
+            files.forEach { it.cleanup() }
```
henrytao-me, line 106: We can clean up the file as soon as we feed them into the framework.

henrytao-me:
A last follow up PR is still needed when Google update framework to return NOT_FOUND. Then, we can add lastCheckPeriod back.

PR #569, opened 2020-07-09
Add sentry logging

henrytao-me:


fastlane/Fastfile
```diff
@@ -32,7 +32,11 @@ end
 platform :ios do
   desc "Submit a new Covid Alert beta build to Apple TestFlight"
   lane :beta do
-    ensure_git_branch
+
+    increment_build_number(
```
smcmurtry, line 36: My understanding is we don't want to do this unless we can get it working for both iOS and Android. We want to keep the build numbers in sync for both platforms. @dsamojlenko correct me if I'm wrong.

timarney:
We won't be merging this - this is only for debug testing.

PR #572, opened 2020-07-09
Attempt to fix push notification

henrytao-me:
This PR attempts to fix push notification. 

### Initial state is incorrect:
Since we start persisting exposure status, we split `start` into `start` and `init`. Because the initial state is retrieved from local storage, it is async. So, if the app is running in background (through background job), master just call `updateExposureStatusInBackground` without calling init. Based on this assumption, 

1. On Android: ✅  
We call `init` before calling `updateExposureStatusInBackground` which means Android should receive notification. If there is no notification, it could be different issue.

2. On iOS: ❌ 
We don't call `init` in `backgroundScheduler.registerPeriodicTask` which means in some race condition, iOS won't have correct initial state. 

### Logic to show notification:
Also the logic for showing exposed notification is quite fragile to break https://github.com/cds-snc/covid-shield-mobile/blob/master/src/services/ExposureNotificationService/ExposureNotificationService.ts#L151. Essentially, we want to send exposure notification to user only once but the way we are doing it is checking the status from `monitoring` to `exposed`. 

This will break when: If user finishes onboarding and minimize the app, the app still runs in but without `updateExposureStatusInBackground`. It means if the status is changed to `exposed`, the next time `updateExposureStatusInBackground` is run, it won't show notification. In this case, user doesn't know that the app is set to exposed already. 

Ref https://github.com/cds-snc/covid-shield-mobile/issues/570

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,7 +31,6 @@ BackgroundScheduler.registerAndroidHeadlessPeriodicTask(async () => {
     SecureStorage,
     ExposureNotification,
   );
-  await exposureNotificationService.init();
   await exposureNotificationService.updateExposureStatusInBackground();
 });
 
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -42,6 +42,7 @@ export type ExposureStatus =
   | {
       type: 'exposed';
       summary: ExposureSummary;
+      notificationSent?: boolean;
       lastChecked?: {
         period: number;
         timestamp: number;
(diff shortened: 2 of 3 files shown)


src/index.ts
```diff
@@ -31,7 +31,6 @@ BackgroundScheduler.registerAndroidHeadlessPeriodicTask(async () => {
     SecureStorage,
     ExposureNotification,
   );
-  await exposureNotificationService.init();
```
henrytao-me, line 34: This is moved to `updateExposureStatusInBackground `

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -145,14 +141,17 @@ export class ExposureNotificationService {
   }
 
   async updateExposureStatusInBackground() {
-    const lastStatus = this.exposureStatus.get();
+    await this.init();
     await this.updateExposureStatus();
     const currentStatus = this.exposureStatus.get();
-    if (lastStatus.type === 'monitoring' && currentStatus.type === 'exposed') {
+    if (currentStatus.type === 'exposed' && !currentStatus.notificationSent) {
```
henrytao-me, line 147: Change logic of showing notification. This making sure that we show exposed notification once.

PR #577, opened 2020-07-09
Minor changes to language around anonymity

katewilhelm:
I think I need a dev to create a new string on the instructions for entering a number screen. Needs a Body4 I think?

Screen Shot 2020-07-09 at 2 57 42 PM


katewilhelm:
oh also tagging @EliseKa for French

PR #578, opened 2020-07-09
Onboarding stepper position

amazingphilippe:
2 design changes for the onboarding:

1. Stepper is above image
2. Images have a bottom border to anchor them in the screen


timarney:
@henrytao-me - mind take a look at this - need to get the auto-focus to move the Step X of X on the next screen.

PR #579, opened 2020-07-09
Fix wrong hook usage of StorageService in Android background job

henrytao-me:
Follow up of https://github.com/cds-snc/covid-shield-mobile/commit/cf84d735c52a4a9ccb99b0c6df38bce2348fc358

Resolves https://sentry.io/organizations/canadian-digital-service/issues/1775653255/?project=5309250&query=is%3Aunresolved


henrytao-me:
I don't like calling `init` in the constructor as it is async and unpredictable. I will make a change.

Issue #390, opened 2020-07-03
Focus state when pressed doesn't meet contrast


JuliannaR:
**Scenario to reproduce**

Focus state when pressed is unreadable and does not pass contrast min requirements. Focus states should be accessible.

**Expected result**
To have pressed state still be readable for users

**Bug or issue / WCAG**

1.4.3 Contrast minimum
https://www.w3.org/TR/WCAG21/#contrast-minimum

**Screenshots where available**
![Image from iOS (2)](https://user-images.githubusercontent.com/35537308/86165672-a365ff00-bae1-11ea-96e4-87983db130bc.png)

**Device + OS**

iPhone 8 | 13.5.1

amazingphilippe:
Is it possible to design custom states for buttons in react native?

timarney:
We can try using
https://reactnative.dev/docs/touchablehighlight

amazingphilippe:
> touchablehighlight

Looking into this, touchable highlight is also limited by opacity. We can't control individual color values of text and background. 

Flagging that in react-native 0.63, the `Pressable`component seems like it can do wonders. @timarney 
https://reactnative.dev/docs/pressable

mtropper:
If this is an acceptable amount of fade, I can apply it in a matter of minutes.

![image](https://user-images.githubusercontent.com/52504483/87089252-25ea6f00-c204-11ea-8b62-8c6ef8257e52.png)

JuliannaR:
@mtropper @amazingphilippe yes, just enough contrast! 
<img width="716" alt="Screen Shot 2020-07-09 at 5 12 22 PM" src="https://user-images.githubusercontent.com/35537308/87091301-70211f80-c207-11ea-87a2-a37e7fe80a43.png">

PR #588, opened 2020-07-09
increase opacity of touchable buttons

mtropper:
Issue #390 Focus state when pressed doesn't meet contrast #390


timarney:
Looks good - just the merge conflict now.

PR #610, opened 2020-07-10
Test solution for EN popup

### LINKED ISSUE #519: "Turn on exposure notifications" tile is not redirecting when "Push notifications are off" is in the menu
Environment : 
Pixel 3a, Android 10, v25 from Play Store

Scenario : 
- Install app but click on "Cancel" when prompted to turn on EN
- Home screen should have the "COVID Alert is off" screen
- Now go to your phone settings and turn off push notifications for the app
- Re-open the app
- Open menu, you should see a tile "Turn on EN" and a tile "Push notifications are off"
- Click on "Turn on EN"

Expected : 
A popup should be displayed to propose to the user to turn on EN

Issue : 
Nothing happens. The popup is not triggered.
Note : this happens only when both tiles are displayed at the same time. Push notifications tile is redirecting as expected.

Screenshot (Jul  7, 2020 19_41_29)

nickmly:
Testing a solution to see if it will fix #519

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -118,7 +118,6 @@ export class ExposureNotificationService {
     if (this.starting) {
       return;
     }
-    this.starting = true;
 
     await this.init();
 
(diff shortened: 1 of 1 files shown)


henrytao-me:
I don't think this is a related fix. 🤔

PR #617, opened 2020-07-10
Font fix for Android

amazingphilippe:
Using `'notosans'` instead of `'Noto Sans'` makes android display the correct font. Doesn't break on iOS.


smcmurtry:
I'm not seeing a difference when I test this locally in Android. Could you post some screenshots of what you're seeing?
Before:
Screen Shot 2020-07-10 at 2 03 16 PM

After:
Screen Shot 2020-07-10 at 2 03 47 PM

amazingphilippe:
Had to rename android assets. 

Look of the `I` in COVID

Capture d’écran, le 2020-07-13 à 10 17 48

PR #640, opened 2020-07-12
Add HeadlessJsTaskWorker

henrytao-me:
This PR attempts to fix race condition when using BackgroundFetch with Broadcast Receiver. The goal is that when Exposure Notification Service sends summary back to the app via Broadcast Receiver, if there is matchedKeysCount, the app should show exposed notification immediately without waiting for next cycle for Background Fetch to run. 

Resoles https://github.com/cds-snc/covid-shield-mobile/issues/595

## How is it done?
The main idea is that when the app receives summary from Broadcast Receiver, the app should:
- Persist token received from the framework to get ExposureSummary later on.
- Only run React Native JS bundle in headless mode when MainActivity is not run. If MainActivity is running, it means there is already a listener from JS layer listening for response. 
- When React Native JS bundle is run, it should read persisted token and perform getExposureSummary.

So, there are two parts: persist token and start React Native JS Bundle while the app is in background.

#### Persist token

- Decide to use SharedPreference as there are not too much information needed to be persisted. 
- Create a `PendingTokenManager` to make sure everything is synchronized. 
- Create `getPendingExposureSummary` in ExposureNotification bridge for Android and back-fill it with `undefined` for iOS. 
- Adjust logic in `updateExposureStatusInBackground` to better handling state changes and showing notification. 

#### Start React Native JS Bundle while the app is in background

- Main challenge: `HeadlessJsTaskService` from React Native doesn't work https://reactnative.dev/docs/headless-js-android.html#example-usage. I receive `Not allowed to start service Intent` exception in the background when Broadcast Receiver is triggered. See logs:

```
2020-07-11 18:12:07.560 21563-21563/com.google.android.apps.exposurenotification E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.google.android.apps.exposurenotification, PID: 21563
    java.lang.RuntimeException: Unable to start receiver app.covidshield.receiver.TestBroadcastReceiver: java.lang.IllegalStateException: Not allowed to start service Intent { cmp=com.google.android.apps.exposurenotification/app.covidshield.receiver.TestBroadcastReceiver$MyTaskService }: app is in background uid UidRecord{26ef2b9 u0a631 RCVR idle change:uncached procs:1 seq(0,0,0)}
        at android.app.ActivityThread.handleReceiver(ActivityThread.java:3797)
        at android.app.ActivityThread.access$1400(ActivityThread.java:219)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1870)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
     Caused by: java.lang.IllegalStateException: Not allowed to start service Intent { cmp=com.google.android.apps.exposurenotification/app.covidshield.receiver.TestBroadcastReceiver$MyTaskService }: app is in background uid UidRecord{26ef2b9 u0a631 RCVR idle change:uncached procs:1 seq(0,0,0)}
        at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1616)
        at android.app.ContextImpl.startService(ContextImpl.java:1571)
        at android.content.ContextWrapper.startService(ContextWrapper.java:669)
        at android.content.ContextWrapper.startService(ContextWrapper.java:669)
        at app.covidshield.receiver.TestBroadcastReceiver.onReceive(TestBroadcastReceiver.kt:23)
```

- `react-native-background-fetch` has `HeadlessTask` which works nicely https://github.com/transistorsoft/react-native-background-fetch/blob/master/android/src/main/java/com/transistorsoft/rnbackgroundfetch/HeadlessTask.java#L24. It also has a util to check if MainActivity is running `BackgroundFetch.getInstance(context).isMainActivityActive` https://github.com/transistorsoft/transistor-background-fetch/blob/master/android/tsbackgroundfetch/src/main/java/com/transistorsoft/tsbackgroundfetch/BackgroundFetch.java#L263. However, I need to edit `build.gradle` here https://github.com/transistorsoft/react-native-background-fetch/blob/master/android/build.gradle#L30 from `implementation` to `api` in order to access that library. It is done via `patch-package` library.

## How to test?

There is no way to trigger this Broadcast Receiver manually `com.google.android.gms.exposurenotification.ACTION_EXPOSURE_STATE_UPDATED`. In order to test this, we need to "borrow" other system events. One of them which works nicely is `android.intent.action.NEW_OUTGOING_CALL`. See other events here https://developer.android.com/guide/components/broadcast-exceptions. 

**TL;DR:** because this test needs to verify when app is terminated, you need to modify some code, add some logs, and call someone (just call and hang up). 

#### Setup

- One way to verify the app run properly while being terminated is ping a url for logging. To do that: 
  + Run any local server. Ex: https://www.npmjs.com/package/http-server
  + `npm install --global http-server && http-server -p 3001 ./`. This will run a local server at port 3001 in current directory. 
  + You then can do `fetch('http://192.168.0.24:3001/some-info-for-logging').catch(() => {})` in the code where you want to log. 

- In `android/app/src/main/AndroidManifest.xml`, change to:
```xml
<manifest ...>

    ...
    <uses-permission android:name="android.permission.PROCESS_OUTGOING_CALLS" />

    <application ...>
        ...
        <receiver
            android:name=".receiver.ExposureNotificationBroadcastReceiver"
            android:exported="true"
            android:permission="com.google.android.gms.nearby.exposurenotification.EXPOSURE_CALLBACK">
            <intent-filter>
                <action android:name="com.google.android.gms.exposurenotification.ACTION_EXPOSURE_STATE_UPDATED" />
                <action android:name="android.intent.action.NEW_OUTGOING_CALL" />
            </intent-filter>
        </receiver>
    </application>
</manifest>
```

- In `android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt`, change to:
```kotlin
            val summaryAndToken = tokens
                .reversed()
                .mapNotNull { token ->
                    val summary = exposureNotificationClient.getExposureSummary(token).await()
                        .toSummary()
                        .takeIf { it.matchedKeyCount > 0 }
                    summary?.let { Pair(it, token) }
                }
                .firstOrNull() ?: Pair(Summary(
                    daysSinceLastExposure = 1,
                    matchedKeyCount = 1,
                    maximumRiskScore = 10
                ), "test")

            log("getPendingExposureSummary", mapOf(
                "summary" to summaryAndToken?.first,
                "token" to summaryAndToken?.second
            ))
```

- In `android/app/src/main/java/app/covidshield/receiver/ExposureNotificationBroadcastReceiver.kt`, change to:
```kotlin
    override fun onReceive(context: Context, intent: Intent) {
        val action = intent.action
        log("onReceive", mapOf("action" to action))

        if (action == Intent.ACTION_NEW_OUTGOING_CALL) {
            if (shouldStartHeadlessJsTask(context)) {
                startHeadlessJsTaskWorker(context, "")
            } else {
                log("Noop when MainActivity is active")
            }
            return
        }

        if (action == ACTION_EXPOSURE_STATE_UPDATED) {
```

- In `src/shared/log.ts`, add below code with your IP and port:
```ts
export const localLog = async (message: string, params: {[key in string]: any} = {}) => {
  const url = `http://192.168.0.24:3001/${encodeURI(message)}/${encodeURI(JSON.stringify(params))}`;
  await fetch(url).catch(() => {});
};
```

- In `src/services/ExposureNotificationService/ExposureNotificationService.ts`, change to:
```ts
  private async performExposureStatusUpdate(): Promise<void> {
    ...
    try {
      localLog('performExposureStatusUpdate', {keysFileUrls});
      const summary = await this.exposureNotification.detectExposure(exposureConfiguration, keysFileUrls);
      localLog('summary', {summary});
      if (summary.matchedKeyCount > 0) {
    ...
  }

    const summary = await this.exposureNotification.getPendingExposureSummary();
    if (exposureStatus.type === 'diagnosed' || (summary?.matchedKeyCount || 0) <= 0) {
      return;
    }
    const today = new Date();
    this.exposureStatus.append({
      type: 'exposed',
      summary,
      lastChecked: {
        timestamp: today.getTime(),
        period: periodSinceEpoch(today, HOURS_PER_PERIOD),
      },
      notificationSent: true, // << add this line to always show notification when testing
    });
  }
```

#### Run

- Run `yarn && yarn start`.
- Build the app from Android Studio to test device. 
- Pass onboarding screen.
- Open AppInfo (long press on app icon) > Permissions > Allow `Call logs` because we use outgoing call system event for simulation.
- Open logcat in Android Studio.
- Minimize the app.
- Make a call to someone and hang up. 
- Verify that `android.intent.action.NEW_OUTGOING_CALL` is received and `Noop when MainActivity is active`
```
2020-07-11 21:41:39.200 26435-26435/? D/CovidShield ExposureNotificationBroadcastReceiver: onReceive {action=android.intent.action.NEW_OUTGOING_CALL}
2020-07-11 21:41:39.203 26435-26435/? D/CovidShield ExposureNotificationBroadcastReceiver: Noop when MainActivity is active {}
```
- Terminate the app. 
- Make a call to someone and hang up. 
- Verify exposed notification show up and the log looks like below:
```
2020-07-12 12:04:25.630 23091-23091/com.google.android.apps.exposurenotification D/CovidShield ExposureNotificationBroadcastReceiver: onReceive {action=android.intent.action.NEW_OUTGOING_CALL}
2020-07-12 12:04:25.630 23091-23091/com.google.android.apps.exposurenotification D/CovidShield ExposureNotificationBroadcastReceiver: startHeadlessJsTaskWorker {token=}
2020-07-12 12:04:26.372 23091-23305/com.google.android.apps.exposurenotification D/CovidShield HeadlessJsTaskWorker: HeadlessJsTaskWorker {shouldStartHeadlessJsTask=true}
2020-07-12 12:04:26.382 23091-23091/com.google.android.apps.exposurenotification D/CovidShield CustomHeadlessTask: onHeadlessJsTaskStart {}
2020-07-12 12:04:29.799 23091-23305/com.google.android.apps.exposurenotification D/CovidShield StandaloneCoroutine: getPendingExposureSummary {summary=Summary(daysSinceLastExposure=1, matchedKeyCount=1, maximumRiskScore=10), token=test}
2020-07-12 12:04:30.274 23091-23091/com.google.android.apps.exposurenotification D/CovidShield CustomHeadlessTask: onHeadlessJsTaskFinish {}
```
- Also verify the local server log that you setup earlier to see if they are hit.
- Play around with multiple calls and verify it works as expected.


android/app/src/main/java/app/covidshield/utils/PendingTokenManager.kt
```diff
@@ -0,0 +1,44 @@
+package app.covidshield.utils
+
+import android.annotation.SuppressLint
+import android.content.Context
+import app.covidshield.MainApplication
+
+private const val TOKEN = "TOKEN"
```
timarney, line 7: Should get set via an .env var?

henrytao-me: This is just RANDOM string. No need to use .env.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -371,4 +381,22 @@ export class ExposureNotificationService {
...
+    }
+    const today = new Date();
+    this.exposureStatus.append({
+      type: 'exposed',
+      summary,
+      lastChecked: {
+        timestamp: today.getTime(),
+        period: periodSinceEpoch(today, HOURS_PER_PERIOD),
+      },
+      notificationSent: false,
```
henrytao-me, line 399: This should be removed

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -371,4 +381,22 @@ export class ExposureNotificationService {
...
+    const exposureStatus = this.exposureStatus.get();
+    if (exposureStatus.type === 'diagnosed' || (summary?.matchedKeyCount || 0) <= 0) {
+      return;
+    }
+    const today = new Date();
+    this.exposureStatus.append({
+      type: 'exposed',
+      summary,
+      lastChecked: {
+        timestamp: today.getTime(),
```
henrytao-me, line 396: I assume the broadcast receiver from the framework won't have more than day of delay. So, when we receive data from broadcast receiver, we could assume lastChecked is today.

android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
```diff
@@ -121,6 +122,33 @@ class ExposureNotificationModule(context: ReactApplicationContext) : ReactContex
         }
     }
 
+    @ReactMethod
+    fun getPendingExposureSummary(promise: Promise) {
+        promise.launch(this) {
+            val tokens = PendingTokenManager.instance.retrieve()
+            PendingTokenManager.instance.clear()
+
+            val summaryAndToken = tokens
```
henrytao-me, line 131: If there is more than one token is pending, we could assume that one token has matchedKeyCount is enough to say people is exposed.

PR #641, opened 2020-07-12
Validating exposure configuration against a JSON Schema.

jeberhardt:
Created a JSON Schema for the exposure configuration, and using the `jsonschema` package to validate the downloaded JSON.


package.json
```diff
@@ -54,6 +55,7 @@
     "reanimated-bottom-sheet": "^1.0.0-alpha.20",
     "rn-faded-scrollview": "^1.0.8",
     "tweetnacl": "^1.0.3",
+    "url": "^0.11.0",
```
timarney, line 58: What are we using this for?

src/services/ExposureNotificationService/ExposureConfigurationValidator.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export class ExposureConfigurationValidator {
+  validateExposureConfiguration(exposureConfiguration: ExposureConfiguration, schema: Schema): ValidatorResult {
+    const validator = new Validator();
+    const validatorResult = validator.validate(exposureConfiguration, schema);
+    if (!validatorResult.valid) {
+      console.log('invalid json');
+      console.log(validatorResult.errors.toString());
+      throw new ExposureConfigurationValidationError(
```
timarney, line 19: Will this cause a fallback to the default?

timarney: Let's add some test cases for some "bad" configs

package.json
jeberhardt: This is needed by the `jsonschema`. I was surprised when it showed up as missing, since `jsonschema` didn't list any dependencies. Apparently `React` includes the `url` package by default, but `react-native` does not.

src/services/ExposureNotificationService/ExposureConfigurationValidator.ts
jeberhardt: Yes, it will cause a fallback to either using a previously saved or the default config. This thrown error is caught in `performExposureStatusUpdate`, in *ExposureNotificationService.ts*

timarney:
Need to fix up this conflict and add a couple of "bad config" unit tests

jeberhardt:
Messed up a rebase. Closing this PR.

PR #643, opened 2020-07-12
Extract usage of console.log

henrytao-me:
Extract usage of console.log to a util.


src/services/BackgroundSchedulerService/BackgroundSchedulerService.ts
```diff
@@ -17,29 +18,35 @@ const registerPeriodicTask = (task: PeriodicTask) => {
       stopOnTerminate: false,
     },
     async taskId => {
+      captureMessage('runPeriodicTask', {taskId});
       try {
         await task();
-      } catch {
-        // noop
+      } catch (error) {
+        captureException('runPeriodicTask', error);
```
timarney, line 25: We should either log error.message here or JSON.stringify

PR #661, opened 2020-07-13
A11y list update

jelen-HC:
Have built-in a11y list functionality from scratch since there does not seem to be any proper accessibilityRole within React Native to handle this. With the screen reader active, the lists now will properly read "List Start", "List End", as well as properly acknowledge the bullets. The Privacy Policy as been added to the "en" & "fr" JSON content to allow it to work with this new functionality as it was not possible in its current form; due to this I also removed the old privacy policy files from "assets" since they are no longer being used.


nickmly:
This works good on Android with Talkback, however I think the translations are out of sync as most of the Home screens had no text content in them. cc. @timarney

timarney:
@JuliannaR if you can give this a look on VoiceOver and than let's see about breaking this PR up.

Ideally if we can get a PR for the list component pieces - and separate PRs for the content pieces that would work best.

It's possible they might want to pull this back upstream without all our content.

JuliannaR:
@timarney I'll finish setting up xcode in the am, going to run through this with Dave. I've got the .IPA and .APK to start on this.
Ideally if this works pulling out the component will support upstream and if/when other lists pop up we would be able to pull it in.

jelen-HC:
Update the EN and FR json files to match missing content from master to resolve conflicts. Unable to resolve the remaining conflicts myself as the changes are required for the update for the list accessibility to function correctly, and need to be reviewed.

PR #667, opened 2020-07-13
code input styling

amazingphilippe:
- Updates how we style the code input so it looks like the mockpus
- Chunks the code into 2x4 to align with the health portal code generator


timarney:
@amazingphilippe this looks great.

@JuliannaR I gave it a test with Talk Back - seem good.

I'll let you review with Talkback and VoiceOver before we merge and or adjust it.

Nice work Phil :)

PR #668, opened 2020-07-13
Fastlane environment mgmt

dsamojlenko:
## What this does
- Strips secrets from `.env` and `.env.production` and checks them into source control consistent with upstream
- Merges `.env.secrets` with `.env` and `.env.production` at build time
- Makes a bunch of helper lanes private

## Why?
- This brings us into alignment with upstream where .env and .env.production are committed, and have occasionally caused issues when pulling from upstream. 
- Additionally, this gets our version name and build number committed to the Repo which will help with CI/deployment.

## Required changes
- Once merged, you should place secrets (`HMAC_KEY`, `SENTRY_DSN`, `SUBMIT_URL`, `RETRIEVE_URL`, `APP_ID_IOS` and `APP_ID_ANDROID`) into a `.env.secrets` file at the root of the project.
- `.env` and `.env.production` files are now checked into source control, so you can go ahead and delete those locally.


timarney:
@dsamojlenko I'll review this with you today.

Issue #674, opened 2020-07-14
Some area not scroll-able in landscape on iOS and Android


qanitah:
The text seems to be on the left of the screen and there is empty space on the right. If the user tries to use the white space on right of the scrollbar, they're not able to scroll the page up and down. 

Highlighted area in the screenshots is not scroll-able.
Android 10 on a large screen leaves a lot of space on the right. iPhone Plus might have similar issue.

Android 8.1.0 on Samsung Galaxy
Android 10 on OnePlus 7
iOS 13.5.1 on iPhone 7

![image](https://user-images.githubusercontent.com/14280062/87438958-2d23cb00-c5be-11ea-8540-3479f2d2c2cf.png)

![image](https://user-images.githubusercontent.com/14280062/87439589-f601e980-c5be-11ea-82cb-c612b3fb81f4.png)

![image](https://user-images.githubusercontent.com/14280062/87440042-7cb6c680-c5bf-11ea-8424-2bd211d3be5c.png)

sarasuri:
Samsung Galaxy J3 Prime
Android version 7.0
App version: 1.0 (47)

Same issue for both COVID Alert is off/Active screens as mentioned above.

JuliannaR:
Combined landscape orientation issues here https://github.com/cds-snc/covid-shield-mobile/issues/701

PR #677, opened 2020-07-14
maximize width in landscape issue#674

mtropper:
#674 Some area not scroll-able in landscape on iOS and Android #674


mtropper:
@JuliannaR @timarney I've made it so the screen is full screen in both orientations

mtropper:
@JuliannaR I also fixed the Bug @srepussard was having with the rotation lock

srepussard:
> This looks like it works well.

It is still a bug for me on Android (v49) : 
- phone rotation lock is on
- but when I rotate the phone when in the app, the screen rotates

mtropper:
@srepussard it hasn't been merged yet. So not available on releases yet

timarney:
@mtropper nicely done 👍

PR #705, opened 2020-07-15
Fix issue with exposureStatus observable.

jeberhardt:
Fix issue with exposureStatus observable that was throwing an error, and not persisting the exposureStatus.

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -145,44 +145,16 @@ export class ExposureNotificationService {
 
   async updateExposureStatusInBackground() {
     await this.init();
-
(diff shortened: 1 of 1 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -145,44 +145,16 @@ export class ExposureNotificationService {
...
-        });
-      }
-    });
-
     try {
       captureMessage('updateExposureStatusInBackground', {exposureStatus: this.exposureStatus.get()});
       await this.processPendingExposureSummary();
+      await this.processNotification();
       await this.updateExposureStatus();
+      await this.processNotification();
```
timarney, line 153: Duplicate call?

Issue #676, opened 2020-07-14
Update app to use UTC vs device local time


timarney:
We need to standardize the app to use UTC to ensure consistency.

Ideally any date time functions / calls should be here:

```
 src/shared/date-fns.ts
```

timarney:
@dsamojlenko tagging you to mark as "this is assigned" we'll sync on Slack for this.

henrytao-me:
Note: we should continue looking into UTC VS device timezone to find all race conditions that could mess up user experience.  For example: 
- Missing the last key upload because device timezone reaches 0 day and reset the UI while in UTC timezone, user still has to upload one more time.

PR #710, opened 2020-07-15
Code refactor to use helper functions for current time (#676)

pisajew:


src/screens/home/views/DiagnosedView.tsx
```diff
@@ -18,7 +18,7 @@ export const DiagnosedView = ({isBottomSheetExpanded}: {isBottomSheetExpanded: b
 
   if (exposureStatus.type !== 'diagnosed') return null;
 
-  const daysLeft = daysBetween(new Date(), new Date(exposureStatus.cycleEndsAt)) - 1;
+  const daysLeft = daysBetween(getCurrentDate(), new Date(exposureStatus.cycleEndsAt)) - 1;
```
timarney, line 21: We may need to re-visit this part

new Date(exposureStatus.cycleEndsAt)

PR #711, opened 2020-07-15
Using default iOS NSAppTransportSecurity settings for optimal security.

jeberhardt:
The default settings for NSAppTransportSecurity provide everything needed to require HTTPS and prevent arbitrary loads.

Please see the note in the README about how this affects development environments.


timarney:
@obrien-j ☝️ 

We're noting in the Readme - what to do for local dev as needed.  

Defaulting to a Prod setup.

PR #716, opened 2020-07-15
Initial language select

timarney:
closes https://github.com/cds-snc/covid-shield-mobile/issues/700

Screen Shot 2020-07-15 at 8 18 12 PM


timarney:
@JuliannaR @amazingphilippe 

Feel free to adjust this as you see fit.

amazingphilippe:
A few adjustments:

1. Buttons fixed at the bottom
2. It should cut the image on smaller screens, but keep the app name visible. Future illustration will be shorter. Lets keep it like this for now.
3. Used the safeAreaView to make sure we don't overlap on OS UI

PR #717, opened 2020-07-16
Use UTC for calculating needSubmission

henrytao-me:
See comment https://github.com/cds-snc/covid-shield-mobile/issues/676#issuecomment-659095038

Ref https://github.com/cds-snc/covid-shield-mobile/issues/676

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -4,15 +4,7 @@ import ExposureNotification, {
   ExposureConfiguration,
 } from 'bridge/ExposureNotification';
 import PushNotification from 'bridge/PushNotification';
-import {
-  addDays,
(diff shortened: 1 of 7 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -230,7 +222,7 @@ export class ExposureNotificationService {
   private async recordKeySubmission() {
     const currentStatus = this.exposureStatus.get();
     if (currentStatus.type !== 'diagnosed') return;
-    this.exposureStatus.append({needsSubmission: false, submissionLastCompletedAt: getMillisSinceUTCEpoch()});
```
henrytao-me, line 233: I found `getMillisSinceUTCEpoch` is confusing. `.getTime()` already returns UTC. So, there is no need for it.

Issue #701, opened 2020-07-15
Landscape orientation related issues (text position, button, and scroll related issues)


JuliannaR:
Text gets cut off in landscape orientation
https://github.com/cds-snc/covid-shield-mobile/issues/699

Back button is missing in landscape orientation #698
https://github.com/cds-snc/covid-shield-mobile/issues/698

Some area not scroll-able in landscape on iOS and Android #674
https://github.com/cds-snc/covid-shield-mobile/issues/674 (in an PR https://github.com/cds-snc/covid-shield-mobile/pull/677 awaiting verification)

amazingphilippe:
#698 is an iOS issue. The "Back to COVID Alert" button is provided by the OS, not the app, not the Ontario page.

PR #725, opened 2020-07-16
Fixed the padding in portrait mode

mtropper:
#701 fixed the extra padding that was added in Portrait mode

### DIFF
--- a/src/screens/home/HomeScreen.tsx
+++ b/src/screens/home/HomeScreen.tsx
@@ -194,6 +194,7 @@ export const HomeScreen = () => {
         <Box
           flex={1}
           paddingTop="m"
+          paddingBottom="m"
           alignSelf="stretch"
           accessibilityElementsHidden={isBottomSheetExpanded}
           importantForAccessibility={isBottomSheetExpanded ? 'no-hide-descendants' : undefined}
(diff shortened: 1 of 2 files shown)


src/screens/home/HomeScreen.tsx
```diff
@@ -193,7 +193,7 @@ export const HomeScreen = () => {
       <Box flex={1} alignItems="center" backgroundColor={strToBackgroundColor(backgroundColor)}>
         <Box
           flex={1}
-          padding="m"
+          paddingTop="m"
```
henrytao-me, line 196: It's better to add `paddingBottom="m"` too.

mtropper:
added paddingBottom as well :)

amazingphilippe:
I can also see that we have extra padding in OverlayView.tsx. 
line 153. we can just remove the two padding attributes!

Simulator Screen Shot - iPhone 11 - 2020-07-16 at 15 41 49

mtropper:
@amazingphilippe the padding was added because in landscape part of the text gets covered. 
image

mtropper:
@JuliannaR @amazingphilippe I modified it so the padding only appears in Landscape mode

PR #738, opened 2020-07-16
Fix for TouchableOpacity not working on Android in production

arrygoo:
Currently if you set `TEST_MODE=false` in `.env`, TouchableOpacity on Android will stop receiving `onPress` events. 

This seems to be due to nesting native stack navigators, very similar to what's described at:
https://github.com/react-navigation/react-navigation/issues/6691

This PR changes the inner StackNavigator to use `react-navigaton`'s , which still uses native screens, and fixes the issue above.


timarney:
Failing CI -> `stackAnimation' does not exist in type`

timarney:
https://github.com/cds-snc/covid-shield-mobile/issues/755

PR #781, opened 2020-07-18
One-time-code allow over 8-10 characters

timarney:
Fixes a bug where the input would allow for 10 characters but the submit button would disable after 8 characters.


jeffmaher:
Taking a look now.

jeffmaher:
Works on iOS. Testing on Android now.

jeffmaher:
Works on Android.

PR #787, opened 2020-07-19
Removes bottom drawer spacer

timarney:
**Before:**

**After:**


jeffmaher:
Looks good on iOS and Android!

PR #806, opened 2020-07-20
Automatic dashes added in CodeInput

nickmly:
**WIP do not merge**

Automatically format code input like so:
`aaa-bbb-cccc`


timarney:
@nickmly thinking we should close this and we can re-visit later

PR #811, opened 2020-07-21
Prioritize expsoure summary by using lastExposureTimestamp

henrytao-me:
This PR prioritize exposure summary that has larger `lastExposureTimestamp`. See more detail in the issue.

Resolves https://github.com/cds-snc/covid-shield-mobile/issues/803

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -340,12 +340,13 @@ export class ExposureNotificationService {
         return finalize();
       }
       return finalize({needsSubmission: await this.calculateNeedsSubmission()});
-    } else if (
-      currentStatus.type === 'exposed' &&
-      currentStatus.summary.daysSinceLastExposure >= EXPOSURE_NOTIFICATION_CYCLE
-    ) {
-      this.exposureStatus.set({type: 'monitoring', lastChecked: currentStatus.lastChecked});
-      return finalize();
+    } else if (currentStatus.type === 'exposed') {
+      const today = getCurrentDate();
+      const lastExposureAt = new Date(currentStatus.summary.lastExposureTimestamp || today.getTime());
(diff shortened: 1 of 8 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -382,20 +382,28 @@ export class ExposureNotificationService {
   private async processPendingExposureSummary() {
     const summary = await this.exposureNotification.getPendingExposureSummary();
     const exposureStatus = this.exposureStatus.get();
-    if (exposureStatus.type === 'diagnosed' || (summary?.matchedKeyCount || 0) <= 0) {
+    if (exposureStatus.type === 'diagnosed' || !summary || summary.matchedKeyCount <= 0) {
```
smcmurtry, line 385: much more readable!

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -340,12 +340,13 @@ export class ExposureNotificationService {
...
       return finalize({needsSubmission: await this.calculateNeedsSubmission()});
-    } else if (
-      currentStatus.type === 'exposed' &&
-      currentStatus.summary.daysSinceLastExposure >= EXPOSURE_NOTIFICATION_CYCLE
-    ) {
-      this.exposureStatus.set({type: 'monitoring', lastChecked: currentStatus.lastChecked});
-      return finalize();
+    } else if (currentStatus.type === 'exposed') {
+      const today = getCurrentDate();
+      const lastExposureAt = new Date(currentStatus.summary.lastExposureTimestamp || today.getTime());
```
henrytao-me, line 345: The `||` here is just a safety check, just in case people upgrades from previous build. So that, it won't crash. If people is in exposed state in previous build, they will be in exposed state forever even after upgrading to this build until they are being exposed again.

PR #801, opened 2020-07-20
Landscape issues with talkback

nickmly:
Fixes #704 

- When clicking 'Learn more about how it works' button in Step 3 of Onboarding and then closing that screen, focus will be directed at the button, rather than the top of the screen (step indicator)
- Same thing for Step 5 with the 'Learn more about privacy' button

Tested on Android with Talkback using portrait and landscape modes.

 **Note: needs to be tested on iOS with VoiceOver**

Issue #755, opened 2020-07-17
TEST_MODE - false causing issues with TouchableOpacity on Android


timarney:
Currently if you set TEST_MODE=false in .env, TouchableOpacity on Android will stop receiving onPress events.

Ref: https://github.com/cds-snc/covid-shield-mobile/pull/738

PR #817, opened 2020-07-21
Fix screenreader focus issues on page change by switching to old navigator

### LINKED ISSUE #799: Page continues to read the previous/closed page image
With VoiceOver on, when the user is going item by item from How it works (step 3 of 6) to "Learn more about how it works" and comes back to step 3 by either clicking Close or by going through all 4 of it's tabs, it announces "Hand drawn illustration of mobile phone in a pocket. sending and receiving signals". It announces first when the user gets back and then continues to announce this when the user swipes right to read next item.

This is also the case for Step 5 for "Learn more about privacy". VoiceOver continues to announce "Privacy policy heading".

Please see video of the Step 3. I'm only swiping right and left. https://photos.app.goo.gl/TCFbUyq7SPJQM1WZ6

iOS 13.5.1 on iPhone 7

smcmurtry:
Closes #799, closes #801

Switch to the old navigator, keep the same nav nesting if `TEST_MODE === false`, just disable the gesture that opens the drawer in that case. Still needs testing on Android to make sure this will not re-introduce #755.

### DIFF
--- a/src/testMode/DemoMode.tsx
+++ b/src/testMode/DemoMode.tsx
@@ -10,6 +10,7 @@ import {
   useReportDiagnosis,
 } from 'services/ExposureNotificationService';
 import {captureMessage} from 'shared/log';
+import {TEST_MODE} from 'env';
 
 import {RadioButton} from './components/RadioButtons';
 import {MockProvider} from './MockProvider';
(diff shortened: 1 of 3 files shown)


src/testMode/DemoMode.tsx
```diff
@@ -162,10 +163,11 @@ export const DemoMode = ({children}: DemoModeProps) => {
     return Component;
   }, [children]);
 
+  // disable the drawer if TEST_MODE === false
   return (
     <MockProvider>
```
arrygoo, line 168: Should `<MockProvider>` also be disabled when in `TEST_MODE===true`?

PR #823, opened 2020-07-21
Fixes blank tutorial pages when in landscape

joshpc:
# What this does
This fixes an issue where tutorial pages would render as blank when presented in landscape mode.

Screen Shot 2020-07-21 at 1 51 16 PM

# Why? How?
The `Carousel`'s rendering optimization (of not rendering off-screen elements) was not rendering the first element when the Carousel itself was completely offscreen. A re-render was properly triggering it to be visible, but the effect is a little jarring (it's a screen flicker) or broken (completely blank in landscape mode as no re-render is triggered.)

By disabling `removeClippedSubviews` this forces the views to be rendered and visible. A side-effect of this is that the views that are off-screen are rendered and on the stack, but given the small quantity of views, this may be an acceptable change.

An alternative fix would be to use `Carousel`'s `triggerRenderingHack`-- see: https://github.com/archriss/react-native-snap-carousel/blob/master/doc/PROPS_METHODS_AND_GETTERS.md#methods -- which addresses this issue specifically.

**Note:** This issue was not evident when the device was in portrait mode as a weird "jiggle" is present when opening the view in portrait mode. That jiggle seems to be a rendering bug. This appears to force the view to re-render (correctly.) Gif with slow animations below:

**Without fix**
broken-portrait
broken-landscape

**With fix**
fixed-portrait
fixed-landscape

# Tested
- Tested on iOS Simulators and iPhone X

### DIFF
--- a/src/screens/tutorial/Tutorial.tsx
+++ b/src/screens/tutorial/Tutorial.tsx
@@ -62,6 +62,7 @@ export const TutorialScreen = () => {
             onSnapToItem={newIndex => setCurrentStep(newIndex)}
             importantForAccessibility="no"
             accessible={false}
+            removeClippedSubviews={false}
           />
         </View>
         <Box flexDirection="row" borderTopWidth={2} borderTopColor="gray5">


timarney:
Thanks @joshpc  we'll take a look as soon as we can.

timarney:
Nice :)

Issue #800, opened 2020-07-20
Page doesn't scroll up for the very bottom buttons automatically (build 58)


qanitah:
With VoiceOver on, when the user is at the bottom of these two onboarding screens (Step 3 and Step 5), the page doesn't automatically scroll up to show these clearly. Since there is nothing after this text, the next item is "Back" button. The text visible on the screen ends up with failed colour contrast since it's blurred/greyed out (cause it's at the end of the text area).

Please see screen shots.

iOS 13.5.1 on iPhone 7

![image](https://user-images.githubusercontent.com/14280062/87974912-d07e4f80-ca98-11ea-9d0c-f0fa6f5ad30e.png)

![image](https://user-images.githubusercontent.com/14280062/87974880-bd6b7f80-ca98-11ea-884b-196c705571ec.png)

JuliannaR:
@qanitah this is a visual fade and when the screen continues to scroll downward the colour of this changes to be dark and readable. When the element is focused it meets contrast. This is a visual cue that there is content below the fold.

qanitah:
Yes it's the fade for the bottom of the screen but page doesn't automatically scroll up. The next item is Back button. It never becomes actual colours with VoiceOver on since there's nothing below this. It's also for the Provinces screen (Step 6).

timarney:
Can the bottom of these pages have some additional padding added to help with the issue?

JuliannaR:
@timarney it would be a workaround but would help.

timarney:
Closing this needs to be retested on latest versions.

PR #838, opened 2020-07-22
Issue 800 added padding in bottom

mtropper:
Issue #800 
* Added padding in bottom so it's less covered by the overlay.

### DIFF
--- a/src/components/BulletPointCheck.tsx
+++ b/src/components/BulletPointCheck.tsx
@@ -10,7 +10,7 @@ export const BulletPointCheck = ({listAccessibile, text}: {listAccessibile: stri
 
   switch (listAccessibile) {
     case 'listStart':
-      bulletLabel = `${i18n.translate(`A11yList.Bullet`)} \n ${i18n.translate(`A11yList.Start`)}`;
+      bulletLabel = `${i18n.translate(`A11yList.Start`)}\n ${i18n.translate(`A11yList.Bullet`)} `;
       break;
 
     case 'listEnd':
@@ -24,7 +24,7 @@ export const BulletPointCheck = ({listAccessibile, text}: {listAccessibile: stri
   }
 
   return (
-    <Box flexDirection="row" marginBottom="s">
+    <Box accessible flexDirection="row" paddingBottom="m">
       <Box accessible accessibilityLabel={bulletLabel} marginTop="xxs" flex={0}>
         <Icon size={20} name="icon-green-check" />
       </Box>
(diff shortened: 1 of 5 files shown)


henrytao-me:
Good catch @smcmurtry 👍

mtropper:
@smcmurtry in IOS normally the bullet and the text are read together but looks like android doesn't follow that rule. I feel its redundant to swipe twice to get the content of each bullet. But I guess its a personal preference. I'll revert it back. 
Screenshot 2020-07-23 at 1 30 13 PM

mtropper:
@smcmurtry when I was looking at your comment earlier, I could only see the still image. Now I can see the animation. That does seems weird. Locally it just reads the bullet and the text together. I will remove that

mtropper:
> I don't think we actually want the `accessible accessibilityRole="button"` props on the `Box` component. The `ButtonSingleLine` component already has the `accessibilityRole= "button"` prop ([ref](https://github.com/cds-snc/covid-shield-mobile/blob/master/src/components/ButtonSingleLine.tsx#L117)). And touchable elements like the button have the `accessible` prop set to true by default.
> 
> Sorry I didn't catch this on my 1st review!

@smcmurtry the tag was added to the box so the focus would go to both the button and the padding below the button so talk back and Voiceover to scroll the button higher up so its not covered by the overly. Otherwise it will not scroll enough for the button to not be under the overlay.

timarney:
> Can the bottom of these pages have some additional padding added to help with the issue?

My original comment was to see if we could add to the bottom of the **pages.**  

I haven't checked out this PR but looks like padding was added to the button .

@amazingphilippe should comment before anything gets merged here.

mtropper:
> > Can the bottom of these pages have some additional padding added to help with the issue?
> 
> My original comment was to see if we could add to the bottom of the **pages.**
> 
> I haven't checked out this PR but looks like padding was added to the button .
> 
> @amazingphilippe should comment before anything gets merged here.

@timarney  The problem isn't with there not being enough padding in the bottom. The problem is that the screen reader doesn't see the overlay at the bottom. So as soon as the button is at the bottom of the page it stops scrolling which puts the button behind the overlay.

amazingphilippe:
I don't know if button padding solves the issue in the most graceful way. Is there a way to make voiceover aware of the bottom fade? My worry is about the voiceover outline being larger than what it needs. Furthermore, voice over reads out the item, so I wouldn't worry about readability. @JuliannaR would there be a scenario where a user uses voice over, without listening to the voice? just for navigation?

Issue #837, opened 2020-07-22
10 char code field allows for more than 10 char


srepussard:
Environment :
Pixel 3a, Android 10, v62

Scenario :
- Navigate to the "Enter a one time key" screen
- Start entering a code until you cannot input any character anymore

Expected :
The app now requires a 10-digit key so the field should only allow for 10 char to be entered

Issue :
The field allows for more than 10 char

![Screenshot (Jul  22, 2020 08_18_01)](https://user-images.githubusercontent.com/47008295/88194866-2a475c80-cbf4-11ea-8f12-48b7be589ac1.png)

timarney:
We'll fix this up - this was a fix last night to allow 

a) small scale beta - which was still 8 digits
b) 10 characters
c) 10 characters with spaces

@emilykuret can you confirm the format + spacing 

XXX XXX XXXX (with spaces) - believe we'll limit to 12 characters total.

srepussard:
Tested on v64 : **Passed**
- max length is now 12 and "Submit" is enabled after min 10
- test cases edited for this

PR #840, opened 2020-07-22
Set CodeInput maxLength to 12

dsamojlenko:
Enforces a limit of 12 chars and a minimum of 10 chars on the CodeInput. #837


src/screens/datasharing/views/FormView.tsx
```diff
@@ -52,8 +52,7 @@ export const FormView = ({value, onChange, onSuccess, onError}: FormViewProps) =
       <Box flex={1} marginHorizontal="m" marginBottom="m">
         <Button
           loading={loading}
-          // @todo update this for 10 digit code
-          disabled={value.length <= 7}
+          disabled={value.length <= 10}
```
timarney, line 55: Believe this should be 
```  
disabled={value.length < 10}
```

PR #844, opened 2020-07-22
Remove special chars in CodeInput

### LINKED ISSUE #839: One time key field allows for special char and more
Environment :
Pixel 3a, Android 10, v62

Scenario :
- Navigate to the "Enter a one time key" screen
- In the key field, enter special characters, even emojis, copy-paste content...

Expected :
The field should only allow alphanumeric characters

Issue : 
The field currently allows for all types of content to be entered

Screenshot (Jul  22, 2020 08_22_36)

nickmly:
Fixes #839 
Replaces any special characters as user types/copy and pastes text into the CodeInput

In TalkBack on Android, it will announce the deletion as the user types a special char (e.g. if they type a dollar sign it will say 'dollar sign deleted')

### DIFF
--- a/src/components/CodeInput.tsx
+++ b/src/components/CodeInput.tsx
@@ -10,7 +10,14 @@ export interface CodeInputProps {
 }
 
 export const CodeInput = ({value, onChange, accessibilityLabel}: CodeInputProps) => {
-  const onChangeTrimmed = useCallback(text => onChange(text.trim()), [onChange]);
+  const onChangeTrimmed = useCallback(
+    text => {
+      // Remove any character that is not alphanumeric (or a space)
+      const modifiedText = text.replace(/[^a-zA-Z0-9\s]*/g, '');
+      onChange(modifiedText.trim());
+    },
+    [onChange],
+  );
   const [isFocus, setIsFocus] = useState(false);
   const onFocus = useCallback(() => setIsFocus(true), []);
   const onBlur = useCallback(() => setIsFocus(false), []);


timarney:
1) This is auto removing spaces
2) This is auto removing -

My concern here is someone needs to "sign-off" on what pattern we're looking for here.


https://github.com/cds-snc/covid-shield-mobile/issues/837#issuecomment-662522258

emilykuret:
Format of the codes will be up to 12 characters including spaces.

timarney:
@JuliannaR @smcmurtry 

If you folks can test this one as well - I want to ensure we're not causing issues for anyone inputting codes.

timarney:
On hold

sashimii:
I would also recommend that the decided upon logic/ReGex pattern for sanitizing these key strings should be placed inside of their own utility function, and perhaps placed in a place like [`src/shared`](src/shared). 

The function name should ideally serve to tell us what the code is doing.

timarney:
@nickmly can you make the suggested change https://github.com/cds-snc/covid-alert-app/pull/844#issuecomment-667491424 today?

src/components/CodeInput.tsx
```diff
@@ -10,7 +10,14 @@ export interface CodeInputProps {
 }
 
 export const CodeInput = ({value, onChange, accessibilityLabel}: CodeInputProps) => {
-  const onChangeTrimmed = useCallback(text => onChange(text.trim()), [onChange]);
+  const onChangeTrimmed = useCallback(
```
jeberhardt, line 13: This function is doing more than just trimming the string, the function name `onChangeTrimmed` should be changed to reflect that.

timarney:
Closes this pending further discussion we'll re-open as needed.

PR #847, opened 2020-07-22
Changed accessibilityLabel to accessibiltyHint for CodeInput

### LINKED ISSUE #843: TextView issue for input of one time code
**Google Play Console Firebase report**

This editable TextView has an android:contentDescription. A screen reader may read this attribute instead of the editable content when the user is navigating.

Screen Shot 2020-07-22 at 1 21 11 PM

nickmly:
Fixes #843 

According to: https://support.google.com/accessibility/android/answer/6378120?hl=en

> Defining an android:contentDescription on any EditText or editable TextView may interfere with an accessibility service's ability to describe, navigate, and interact with text that a user enters within the item.

`android:contentDescription` is mapped to `accessibilityLabel` on the react native layer, so I changed it to a hint to prevent this warning from coming up

**Note: This needs to be tested in Google Play Test Lab**


timarney:
@nickmly let me see if I can get a link for testing with.

PR #849, opened 2020-07-22
Minor change to privacy policy

katewilhelm:
about what happens to exposure logs when you delete the app


lchski:
Just as a note, for anyone curious: this updates the privacy policy to more accurately describe the actual behaviour of the Google/Apple Exposure Notification framework. When you delete an app that uses the framework, the Exposure Logs aren't deleted from the device itself (this is so you can switch between different apps if you so desired). But the Exposure Logs expire after their retention period is up, hence the "after 15 days” addition.

PR #851, opened 2020-07-22
Fix potential null dereference in iOS code

### LINKED ISSUE #789: BVC-150: Possible null deference in GetRandomBytes() function
https://github.com/cds-snc/covid-shield-mobile/blob/611e4d49571568ed83452f0c4246b940c92a7432/ios/CovidShield/CovidShield.m#L16

"The call to malloc isn't checked, so in the event of an OOM situation there will be a null deref at the end of the function when freed."

"Suggestion on new code could look like this:" -- rec's from MOB-M-6 (bbry earlier mem leak finding)

NSMutableData *buff = [NSMutableData dataWithLength:size];
int status = SecRandomCopyBytes(kSecRandomDefault, size, buff.mutableBytes); if (status == 0) {
NSString *base64encoded = [buff base64EncodedStringWithOptions:0];
resolve(base64encoded); }

You can also just check the return value, but honestly, at this point, I'd just recommend going with the higher level wrapper and be done with it.

pisajew:
To fix #789 . On contrary to the original comment this uses low-level API, as NSMutableData could also fail returning nil, so it wouldn't rather save a lot of lines.

### DIFF
--- a/ios/CovidShield/CovidShield.m
+++ b/ios/CovidShield/CovidShield.m
@@ -13,16 +13,17 @@ RCT_EXPORT_MODULE();
 
 RCT_REMAP_METHOD(getRandomBytes, randomBytesWithSize:(NSUInteger)size withResolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
 {
+  int status = errSecSuccess;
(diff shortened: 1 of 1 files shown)


timarney:
@jeberhardt any comments

Believe this covers the issue.

PR #852, opened 2020-07-22
Adding borders around interactable boundaries

### LINKED ISSUE #842: Non text contrast issues related to nav elements, arrow for expanding menu, and thin line for separating element for back/next
**Google Play Console Firebase Accessibility finding:**

The item's text contrast ratio is 1.16. This ratio is based on an estimated foreground color of #FFFFFF and an estimated background color of #EEEEEE. Consider using colors that result in a contrast ratio greater than 4.50 for small text, or 3.00 for large text.

Issue:
Screen Shot 2020-07-22 at 11 20 37 AM
Screen Shot 2020-07-22 at 11 20 47 AM

**Design consideration:**
Capture d’écran, le 2020-07-22 à 12 16 54

amazingphilippe:
This closes #842 

- Borders around buttons: InfoButton, ButtonSingleLine, InfoShareItem, RegionItem, LanguageSelectItem
- Borders around cards: InfoBlock, Tip
- Borders around the shield
- Border in button variants
- Shield handlebar is darker
- Carousel pagination border is darker


timarney:
@amazingphilippe 

On Android the "border top" on the drawer menu goes away when the inner contents of the menu scroll


Are the rounded corners on the bottom intentional?

amazingphilippe:
>"border top" on the drawer menu goes away

Yes, I don't think we can do anything about it, but you may have an idea:
The bottom sheet comes from react-native-reanimated-bottom-sheet, and it can't take any style properties other than rounded corners. So we have to use the box wrapper inside the sheet. 

> Are the rounded corners on the bottom intentional?
No, good catch. I was meaning to do a bit more testing. I'll put this into draft mode to address the two issues.

amazingphilippe:
Just a quick update on this issue: We can keep it just in case, or if we want a quick fix. We are trying to find a better design solution to make sure we have good contrast without the need for borders around all the things. 

Solutions like :
- dividing menu items with a line instead of separating them in blocks. 
- adjusting card and shield colors to strengthen the contrast between those elements.

timarney:
Up to you @amazingphilippe - we'll do doing another build tomorrow likely around noon (latest).

amazingphilippe:
@timarney You can include this in the next build!

PR #853, opened 2020-07-22
Change react-native package for better secure storage.

jeberhardt:
The previous package for storing secure data `react-native-sensitive-info` did not store the data for Android in the keychain.

Replacing that package with `react-native-secure-key-store` allows for both iOS and Android data to be stored in the keychain.

Along with changing the package, I changed the way the **Exposure Configuration** is being stored. It is no longer being stored in the secure storage, but rather just in the AsyncStorage. The exposure configuration is not sensitive data, as it's the same file downloaded from the server for every person.

### DIFF
--- a/android/settings.gradle
+++ b/android/settings.gradle
@@ -3,8 +3,6 @@ include ':react-native-localize'
 project(':react-native-localize').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-localize/android')
 include ':react-native-system-setting'
 project(':react-native-system-setting').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-system-setting/android')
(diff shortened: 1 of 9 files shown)


jeberhardt:
The configuration file is the item in question being downloaded and stored in "non-secure" storage.
-- this file is the same for all users, so there is no need to store it securely
-- the file operations on the downloaded configuration data (write / read) can happen while the app is running in the background and / or the phone is locked, which would mean that the data wouldn't be stored securely, even though it was in the keychain.
-- the code would be different anyway, given point #2.
-- I think it's good to be clear that the configuration isn't stored securely

obrien-j:
Thanks for the context @jeberhardt , lgtm.

PR #855, opened 2020-07-23
Adds a KeyboardAvoidingView to automatically adjust padding/height when keyboard covers content

joshpc:
# What
Enables the user to scroll when the system keyboard covers content on the screen. In this app, it only manifests when a user is entering codes.

# Why? How?
When entering a one-time key in landscape mode, the submit button was obscurred. The user can still press "done" on the native keyboard and it will progress, however the rest of the content is obscurred. This is non-standard behaviour.

This is implemented by using keyboard events, rather than using the `KeyboardAvoidingView`. The react-native KeyboardAvoidingView will create a new view, place all of its content within it, and put a massive `paddingBottom` on it to ensure that the content fits above the keyboard. This uses system provided properties to help manage the keyboard.

This is _fine_ but this means that the content that is scrolled beneath the keyboard is not visible. Additionally, the `KeyboardAvoidingView` causes rendering issues when used within this modal setup.

**Before (static image)**
Screen Shot 2020-07-22 at 12 50 07 PM

** (gif)**
keyboard

# Testing
This was tested on device (iOS, iOS 13.5) and in a simulator.
Tested on an Android emulator (Pixel, Android 10+)

Note:
- It's important to try this in both portrait, landscape, and rotating back and forth between the two modes. The original approach using the `KeyboardAvoidingView` caused render errors (screen would go white on device)

### DIFF
--- a/src/screens/datasharing/DataSharing.tsx
+++ b/src/screens/datasharing/DataSharing.tsx
@@ -1,8 +1,8 @@
-import React, {useCallback, useState} from 'react';
+import React, {useCallback, useState, useEffect} from 'react';
 import {useNavigation} from '@react-navigation/native';
 import {Box, Toolbar} from 'components';
-import {StyleSheet, Alert, ScrollView} from 'react-native';
-import {SafeAreaView} from 'react-native-safe-area-context';
+import {StyleSheet, Alert, ScrollView, Keyboard} from 'react-native';
+import {SafeAreaView, useSafeAreaInsets} from 'react-native-safe-area-context';
 import {useI18n} from 'locale';
 import {useExposureStatus} from 'services/ExposureNotificationService';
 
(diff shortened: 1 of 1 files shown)


timarney:
Android device:


@smcmurtry 

Please try on your devices as well.

My screen above isn't an issue with this PR but the view for landscape is quite different when rotating.  The PR itself looks good.

joshpc:
@timarney I know your team is getting close to release, so if these are too much noise I can hold off -- just trying to pitch in where I can

joshpc:
@henrytao-me the view is scrollable with this PR (the `ScrollView` is there) but there are a few competing interactions. If you scroll down at the "wrong place" right now, the entire view is dismissed despite this being a full screen modal -- is that the behaviour you're seeing?

timarney:
@joshpc definitely not noise :). We'll prioritize them - thanks for the contributions.

timarney:
Closes this pending further discussion we'll re-open as needed.

PR #859, opened 2020-07-23
Force NSURLSession to not cache anything.

jeberhardt:
For added level of security, prevent requests for the diagnosisKeysFile from being cached.


timarney:
@obrien-j any comments?

Issue #857, opened 2020-07-23
Content feedback for onboarding, privacy policy and menu


JuliannaR:
Content suggestions and considerations

<img width="545" alt="Screen Shot 2020-07-23 at 11 11 19 AM" src="https://user-images.githubusercontent.com/35537308/88308226-cb104780-ccda-11ea-99cf-d32543c5788c.png">
<img width="543" alt="Screen Shot 2020-07-23 at 11 12 29 AM" src="https://user-images.githubusercontent.com/35537308/88308245-d1062880-ccda-11ea-838a-e152c4fe9add.png">
<img width="553" alt="Screen Shot 2020-07-23 at 11 13 54 AM" src="https://user-images.githubusercontent.com/35537308/88308270-da8f9080-ccda-11ea-9c1d-41469e9e3a33.png">
<img width="549" alt="Screen Shot 2020-07-23 at 11 15 26 AM" src="https://user-images.githubusercontent.com/35537308/88309908-014ec680-ccdd-11ea-924b-3b68de9b5eb4.png">
<img width="554" alt="Screen Shot 2020-07-23 at 11 11 47 AM" src="https://user-images.githubusercontent.com/35537308/88309923-0a3f9800-ccdd-11ea-9578-25c4fb62e919.png">

@katewilhelm this one is for you to decide on the best approach.

timarney:
@katewilhelm @JuliannaR can you comment and or close this?

timarney:
Going to close this as we can re-open as needed.

PR #860, opened 2020-07-23
Fix bug where iOS system status was not updating quickly enough

### LINKED ISSUE #861: HomeScreen - showing "all set" when Covid Alert is off
The HomeScreen view is showing "all set" status when Covid Alert is off (or in an unknown / undefined state).

smcmurtry:
Closes #861 (Part of #857)

Added logic to set systemStatus to disabled if we have an iOS error indicating the user has disabled the framework. I am just checking against a string right now but I know this is not the correct way to check the type of error. @henrytao-me do you know how to do that properly?

Screen Shot 2020-07-23 at 11 11 47 AM

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -131,6 +131,10 @@ export class ExposureNotificationService {
       await this.exposureNotification.start();
     } catch (error) {
       captureException('Cannot start EN framework', error);
+      if (error.message === 'ENErrorCodeNotAuthorized (User denied)') {
+        this.systemStatus.set(SystemStatus.Disabled);
+        return;
+      }
       this.systemStatus.set(SystemStatus.Unknown);
       return;
     }
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -131,17 +131,11 @@ export class ExposureNotificationService {
       await this.exposureNotification.start();
     } catch (error) {
       captureException('Cannot start EN framework', error);
-      // 4: 'ENErrorCodeNotAuthorized (User denied)'
-      // https://developer.apple.com/documentation/exposurenotification/enerrorcode/enerrorcodenotauthorized?language=objc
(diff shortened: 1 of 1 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -131,6 +131,10 @@ export class ExposureNotificationService {
       await this.exposureNotification.start();
     } catch (error) {
       captureException('Cannot start EN framework', error);
+      if (error.message === 'ENErrorCodeNotAuthorized (User denied)') {
```
timarney, line 134: I think we can use just the "ENErrorCodeNotAuthorized" part of the message or use the error code.  Preferably the error code if possible.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -131,6 +131,12 @@ export class ExposureNotificationService {
       await this.exposureNotification.start();
     } catch (error) {
       captureException('Cannot start EN framework', error);
+      // 4: 'ENErrorCodeNotAuthorized (User denied)'
+      // https://developer.apple.com/documentation/exposurenotification/enerrorcode/enerrorcodenotauthorized?language=objc
+      if (error.code === '4') {
```
timarney, line 136: @henrytao-me we've opted to handle this based on the error code - any comments?

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -132,10 +132,10 @@ export class ExposureNotificationService {
     } catch (error) {
       captureException('Cannot start EN framework', error);
       this.systemStatus.set(SystemStatus.Unknown);
-      return;
     }
 
     await this.updateSystemStatus();
+    console.log('this.systemStatus.get()', this.systemStatus.get());
```
henrytao-me, line 138: Remove this.

timarney: Please :)

smcmurtry: whoops!

PR #866, opened 2020-07-23
Reject promise when attempting to access ENManager without authorization.

jeberhardt:
This will reject the promise when attempting to access the ENManager if authorization has not been granted.

The `reject` will be handled on the `react-native` layer.

### DIFF
--- a/ios/CovidShield/ExposureNotification.m
+++ b/ios/CovidShield/ExposureNotification.m
@@ -83,6 +83,10 @@ RCT_REMAP_METHOD(getStatus, getStatusWithResolver:(RCTPromiseResolveBlock)resolv
 
 RCT_REMAP_METHOD(getTemporaryExposureKeyHistory, getTemporaryExposureKeyHistoryWithResolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
 {
+  if (ENManager.authorizationStatus != ENAuthorizationStatusAuthorized) {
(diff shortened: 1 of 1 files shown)


timarney:
> EN is ON + Bluetooth is OFF

@henrytao-me this feels like a separate problem.

henrytao-me:
IMO, it's same problem and solution. 🤔

timarney:
They end up tied together but for error handling I think this PR should only handle the EN - user has turned off case.

henrytao-me:
TL;DR
I test it again and agree that it is not a problem.

## How to test
- Have EN + Bluetooth ON.
- Enter code time code to be diagnosed. 
- Verify that log without error. 
- Comment out `await this.exposureNotification.start();`
- Terminate and reopen the app. 
- Observe `performExposureStatusUpdate` hangs forever. 

## Explain
- It seems like on iOS we have to call `start` the framework whenever we open the app before doing anything else even EN and Bluetooth are ON in previous open. 
- If user manually disables EN in Settings, the app cannot turn on EN framework programmatically, even with `start`. In contrast, Android can. 
- React Native background job on iOS is also different than Android. When React Native iOS run in the background, it runs the whole app which will result as `start` from the framework is being called. That's why we see exposure check and everything is working. Combining with calling `start` does not shows consent dialog automatically, it's a perfect combination on iOS. In contrast, Android background job doesn't call start and it doesn't need to call start as long as EN is enabled in previous open.

PR #873, opened 2020-07-24
Attempt to check Google Play Services

henrytao-me:
TBD


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -131,10 +131,10 @@ export class ExposureNotificationService {
       await this.exposureNotification.start();
     } catch (error) {
       captureException('Cannot start EN framework', error);
-      this.systemStatus.set(SystemStatus.Unknown);
```
henrytao-me, line 134: This is not necessary anymore as `updateSystemStatus` below will update to proper state.

henrytao-me:
Tested on both phone that have Google Play Services deactivated and up-to-date phone. Works as expected.

PR #884, opened 2020-07-24
Add "Something went wrong" screens

### LINKED ISSUE #872: Handle Google Play Services is not available on device
## What?

Given a phone where the EN Framework is available for a device but Play services isn't updated.

The app can be installed but end up in an incorrect state. 

i.e. indicating the phone is all set / scanning (even with no EN framework on the device)

smcmurtry:
Closes #872. 

This PR adds 2 new screens (see figma for designs). The 1st, `FrameworkUnavailableView`, is to deal w/ the case where android phones have a compatible OS, install the app, but have an old version of Google Play Services that does not support the EN Framework. They will now see `FrameworkUnavailableView`:
Screen Shot 2020-07-23 at 10 43 47 PM

The 2nd screen is the "fall through" screen, that the user will see if their system state is unknown/undefined, since we definitely do not know that the app is monitoring for exposures in this case. `UnknownProblemView`:
Screen Shot 2020-07-23 at 10 43 06 PM


timarney:
We can either handle it here or in another PR but on a device where I receive the "COVID Alert" is not supported I can still see / enter a one-time key.

PR #897, opened 2020-07-24
Announce 'double tap to activate' on external links with Talkback

### LINKED ISSUE #888: Talkback doesn't say double tap to activate for a link
Scenario to reproduce

Samsung Galaxy J3 Prime
Android version 7.0

Samsung S5
Android version 6

App version: Build #64

While Talkback is on, on the “Get a one-time key” screen, Talkback doesn’t say double tap to activate for link "Go to COCID-19 test Results Website”. It announces Go to COCID-19 test Results Website opens in a new window link and then links available you swipe up then right to view. 

Expected result
Talkback should say double tap to activate the link when it announces the link.

nickmly:
Fixes #888 

It looks like the `RectButton` component inside the `Ripple` component is not capable of being both a `link` and having the 'double-tap to activate' hint. Seems like this is only possible with a native `TouchableOpacity` component from react native. 

In order to get around this I've replaced the external links with a `TouchableOpacity` component to announce the proper hint in TalkBack.


timarney:
@smcmurtry please give this a look when you can.

timarney:
@nickmly can you confirm we still need this one i.e. nothing on current master fixed this already.

nickmly:
@timarney Looks like this is still needed. On master with Talkback the link is read without the 'double tap to activate' afterwards. Not sure if this is android specific

Issue #890, opened 2020-07-24
VirtualizedList: You have a large list that is slow to update 


timarney:
We've noticed an edge case on Android (maybe on iOS) 

```
  VirtualizedList: You have a large list that is slow to update - make sure your renderItem function renders components that follow React performance best practices like PureComponent, shouldComponentUpdate, etc. {"contentLength": 2468.571533203125, "dt": 748, "prevDt": 720}
```

During the on boarding I've been able to get this warning when "fast swiping" in the carousel - and reaching the region picker.

Not sure if this is a mounting, unmounting issue or something else.

I've seen some notes regarding the use of FlatList and this issue. 

The impact is the app becomes very sluggish - I haven't seen the app crash in this case but it's possible.

- Need to keep in mind other issues relating to the onboarding flow. 

Will add more refs here:
https://github.com/cds-snc/covid-alert-app/issues/723

PR #899, opened 2020-07-24
Fix VirtualizedList issue

henrytao-me:
This PR fixes VirtualizedList due to props passing to Carousel aren't optimized for rendering. 

Resolve https://github.com/cds-snc/covid-alert-app/issues/890


src/screens/onboarding/Onboarding.tsx
```diff
@@ -21,29 +22,34 @@ export const OnboardingScreen = () => {
   const startExposureNotificationService = useStartExposureNotificationService();
   const isStart = currentStep === 0;
   const isEnd = currentStep === onboardingData.length - 1;
+  const {isScreenReaderEnabled} = useAccessibilityService();
+  const currentStepForRenderItem = isScreenReaderEnabled ? currentStep : -1;
```
henrytao-me, line 26: This should be fixed in follow up PR but we can ship as-is.

PR #909, opened 2020-07-27
Feature/exposure status secure storage

jeberhardt:
Solves the problem of stopping data from being stored in the iOS `/Library` folder, by saving the data in the keychain instead of using React-Native Async storage. 

Setting the `accessible` property to `ACCESSIBLE.ALWAYS_THIS_DEVICE_ONLY` ensures that the data will remain on the phone, and not be backed up to iCloud.

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -193,7 +193,7 @@ export class ExposureNotificationService {
   }
 
   private async init() {
-    const exposureStatus = JSON.parse((await this.secureStorage.get(EXPOSURE_STATUS)) || 'null');
+    const exposureStatus = JSON.parse((await this.secureStorage.get(EXPOSURE_STATUS).catch(() => null)) || 'null');
     this.exposureStatus.append({...exposureStatus});
   }
 
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -196,7 +196,12 @@ export class ExposureNotificationService {
   }
 
   private async init() {
-    const exposureStatus = JSON.parse((await this.secureStorage.get(EXPOSURE_STATUS)) || 'null');
+    let exposureStatus;
+    try {
+      exposureStatus = JSON.parse((await this.secureStorage.get(EXPOSURE_STATUS)) || 'null');
+    } catch (error) {
+      exposureStatus = 'null';
+    }
     this.exposureStatus.append({...exposureStatus});
   }
 
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -8,6 +8,7 @@ import {addDays, periodSinceEpoch, minutesBetween, getCurrentDate, daysBetweenUT
 import {I18n} from 'locale';
 import {Observable, MapObservable} from 'shared/Observable';
 import {captureException, captureMessage} from 'shared/log';
+import {ACCESSIBLE} from 'react-native-secure-key-store';
 
 import {BackendInterface, SubmissionKeySet} from '../BackendService';
 
@@ -106,8 +107,10 @@ export class ExposureNotificationService {
     this.backendInterface = backendInterface;
     this.storage = storage;
     this.secureStorage = secureStorage;
-    this.exposureStatus.observe(status => {
-      this.storage.setItem(EXPOSURE_STATUS, JSON.stringify(status));
+    this.exposureStatus.observe(async status => {
+      await this.secureStorage.set(EXPOSURE_STATUS, JSON.stringify(status), {
+        accessible: ACCESSIBLE.ALWAYS_THIS_DEVICE_ONLY,
+      });
     });
   }
 
@@ -193,7 +196,7 @@ export class ExposureNotificationService {
   }
 
   private async init() {
-    const exposureStatus = JSON.parse((await this.storage.getItem(EXPOSURE_STATUS)) || 'null');
+    const exposureStatus = JSON.parse((await this.secureStorage.get(EXPOSURE_STATUS)) || 'null');
     this.exposureStatus.append({...exposureStatus});
   }
 
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -16,7 +16,7 @@ import exposureConfigurationDefault from './ExposureConfigurationDefault.json';
 import exposureConfigurationSchema from './ExposureConfigurationSchema.json';
 import {ExposureConfigurationValidator, ExposureConfigurationValidationError} from './ExposureConfigurationValidator';
 
-const SUBMISSION_AUTH_KEYS = 'submissionAuthKey';
+export const SUBMISSION_AUTH_KEYS = 'submissionAuthKeys_2';
 const EXPOSURE_CONFIGURATION = 'exposureConfiguration';
 
 export const EXPOSURE_STATUS = 'exposureStatus';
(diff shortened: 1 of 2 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -106,8 +107,10 @@ export class ExposureNotificationService {
     this.backendInterface = backendInterface;
     this.storage = storage;
     this.secureStorage = secureStorage;
-    this.exposureStatus.observe(status => {
-      this.storage.setItem(EXPOSURE_STATUS, JSON.stringify(status));
+    this.exposureStatus.observe(async status => {
+      await this.secureStorage.set(EXPOSURE_STATUS, JSON.stringify(status), {
+        accessible: ACCESSIBLE.ALWAYS_THIS_DEVICE_ONLY,
```
obrien-j, line 112: Just dropping a few links for context and others ability to follow along more easily.

https://developer.apple.com/documentation/security/keychain_services/keychain_items/restricting_keychain_item_accessibility?language=objc

and

https://developer.apple.com/documentation/security/keychain_services/keychain_items/item_attribute_keys_and_values?language=objc#1679100

search "Accessibility Values" as there is no page anchors.

It appears that "kSecAttrAccessibleAlwaysThisDeviceOnly" has been deprecated per -> https://developer.apple.com/documentation/security/ksecattraccessiblealwaysthisdeviceonly?language=objc

I would think we'd want to use kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly as the next option up the list, but appreciate that may introduce a potential problem when trying to get the element on a restart if our background task fires before a user unlocks.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -8,14 +8,15 @@ import {addDays, periodSinceEpoch, minutesBetween, getCurrentDate, daysBetweenUT
...
+import {ACCESSIBLE} from 'react-native-secure-key-store';
 
 import {BackendInterface, SubmissionKeySet} from '../BackendService';
 
 import exposureConfigurationDefault from './ExposureConfigurationDefault.json';
 import exposureConfigurationSchema from './ExposureConfigurationSchema.json';
 import {ExposureConfigurationValidator, ExposureConfigurationValidationError} from './ExposureConfigurationValidator';
 
-const SUBMISSION_AUTH_KEYS = 'submissionAuthKeys';
+const SUBMISSION_AUTH_KEYS = 'submissionAuthKey';
```
obrien-j, line 19: Assuming this is just cosmetic, but figured I'd ask just in case this would potentially cause any issues elsewhere.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -164,11 +167,12 @@ export class ExposureNotificationService {
   async startKeysSubmission(oneTimeCode: string): Promise<void> {
     const keys = await this.backendInterface.claimOneTimeCode(oneTimeCode);
     const serialized = JSON.stringify(keys);
-    console.log(serialized);
     try {
-      await this.secureStorage.set(SUBMISSION_AUTH_KEYS, serialized, {});
+      await this.secureStorage.set(SUBMISSION_AUTH_KEYS, serialized, {
+        accessible: ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY,
+      });
     } catch (error) {
```
obrien-j, line 174: No issues with your catching of the exception, but it did draw my attention to the overall flow of this function.

Can you confirm my interpretation of this?
```

1: Trade OTC for AuthKeys
2: Store AuthKeys into secureStorage
2.5 if 2 throws exception (log, and continue)
3: Put phone (exposureStatus) into 'diagnosed' state

```
If 2.5 is a passthrough (apologies, JS isn't my hometurf), then this would create a pretty faulty condition where the phone was in diagnosed state, and would have no ability to uploadKeys. If the catch ejects from function, no worries, and ignore the thought.

Nothing that needs addressing now, as I'm going to assume this is a very rare situation (if in fact it works the way I thinkg), but figure it's worth flagging.

src/services/ExposureNotificationService/ExposureNotificationService.ts
jeberhardt: Not cosmetic... It was changed so that it wouldn't conflict with any devices that already have the app installed, as changing the accessible options on an existing key doesn't work.

src/services/ExposureNotificationService/ExposureNotificationService.ts
jeberhardt: That's precisely why I choose `ALWAYS_THIS_DEVICE_ONLY`, to handle unknown behaviour when not unlocking after a reboot. Not sure what the best fix is in this case.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -193,7 +197,12 @@ export class ExposureNotificationService {
   }
 
   private async init() {
-    const exposureStatus = JSON.parse((await this.storage.getItem(EXPOSURE_STATUS)) || 'null');
+    let exposureStatus;
+    try {
+      exposureStatus = JSON.parse((await this.secureStorage.get(EXPOSURE_STATUS)) || 'null');
+    } catch (error) {
+      exposureStatus = 'null';
```
henrytao-me, line 204: This should be `null` value, not null string.

henrytao-me: Let's do this without try/catch

```ts
exposureStatus = JSON.parse((await this.secureStorage.get(EXPOSURE_STATUS).catch(() => null)) || 'null');
```

src/services/ExposureNotificationService/ExposureNotificationService.ts
henrytao-me: We don't have any analytics at all on prod. Doing try / catch here will cause the app cannot upload keys in the next cycle. I would prefer to remove try / catch. Set to storage needs to work 100% of the time, I think.

src/services/ExposureNotificationService/ExposureNotificationService.ts
henrytao-me: I understand this part but let's add version instead of removing `s`. Ex: 

```ts
const SUBMISSION_AUTH_KEYS = 'submissionAuthKeys_1';
```

src/services/ExposureNotificationService/ExposureNotificationService.ts
timarney: @henrytao-me go ahead with removing the try catch

src/services/ExposureNotificationService/ExposureNotificationService.ts
timarney: Sounds good let's make that change
```
const SUBMISSION_AUTH_KEYS = 'submissionAuthKeys_1';
```

src/services/ExposureNotificationService/ExposureNotificationService.ts
jeberhardt: > This should be `null` value, not null string.

I was wondering about that, but in the event that line 202 is unable to parse the JSON, it sets it to the same 'null' string.

src/services/ExposureNotificationService/ExposureNotificationService.ts
jeberhardt: The `try/catch` in this case was not added during this PR. Just wondering if it's necessary to fix it at this point, or should it be a new issue?

timarney:
@jeberhardt @henrytao-me should we add a catch here for the `get(SUBMISSION_AUTH_KEYS);`

```javascript
async fetchAndSubmitKeys(): Promise<void> {
    const submissionKeysStr = await this.secureStorage.get(SUBMISSION_AUTH_KEYS);
    if (!submissionKeysStr) {
      throw new Error('No Upload keys found, did you forget to claim one-time code?');
    }
```

henrytao-me:
@timarney No, we don't need to catch it because of two reasons: 
- `fetchAndSubmitKeys` function will only be called after `SUBMISSION_AUTH_KEYS ` has been set. 
- In case, there is something wrong, we already throw exception in the line right below it `No Upload keys found, did you forget to claim one-time code?`. So, `fetchAndSubmitKeys ` will catch both of them in the UI.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -193,7 +193,7 @@ export class ExposureNotificationService {
   }
 
   private async init() {
-    const exposureStatus = JSON.parse((await this.storage.getItem(EXPOSURE_STATUS)) || 'null');
+    const exposureStatus = JSON.parse((await this.secureStorage.get(EXPOSURE_STATUS).catch(() => null)) || 'null');
```
jeberhardt, line 196: @henrytao-me - why is there both null value and null string in this? 

In one case, `exposureStatus = null`, and in the other case, `exposureStatus = 'null'`.

henrytao-me: The bracket is the key. 
- The 1st part is `(await this.secureStorage.get(EXPOSURE_STATUS).catch(() => null))` which run async and return string | null. 
- The 2nd part will make sure it fallbacks to `'null'` string for `JSON.parse`. 

That means `.catch(() => 'null')` will have no different in code execution.

src/services/ExposureNotificationService/ExposureNotificationService.ts
henrytao-me: `JSON.parse('null')` will be `null` value (not string).

timarney:
Closing this in favour of https://github.com/cds-snc/covid-alert-app/pull/918

Issue #820, opened 2020-07-21
iOS BackgroundTask needs requiresNetworkConnectivity


jeberhardt:
## What?

The **BackgroundTasks** framework in iOS has the parameter `requiresNetworkConnectivity` for ** BGProcessingTaskRequest** requests. This means that if the device isn't connected to the network (cellular or WiFi), then the background task will not be triggered, since it would be guaranteed to fail.

The current package being used `react-native-background-fetch` does not allow that parameter to be configured. It is always set to false.

## Suggested Solution
Either update the `react-native-background-fetch` package, or replace it with a package that does support configuring that parameter.

PR #917, opened 2020-07-27
Set flags on BRProcessingTask requests submitted by TSBackgroundFetch

pisajew:
This will ensure that exposure-notification background processing tasks will be submitted
as requiring network connectivity and not needing external power (solves #820 )

### DIFF
--- a/ios/CovidShield/AppDelegate.m
+++ b/ios/CovidShield/AppDelegate.m
@@ -8,6 +8,8 @@
 #import <RNCPushNotificationIOS.h>
 
 #import <TSBackgroundFetch/TSBackgroundFetch.h>
+#import <BackgroundTasks/BackgroundTasks.h>
+#import <objc/runtime.h>
 
 #if DEBUG
 #import <FlipperKit/FlipperClient.h>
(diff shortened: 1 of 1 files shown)


ios/CovidShield/AppDelegate.m
```diff
@@ -124,3 +129,25 @@ - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
 }
 
 @end
+#pragma mark Patch for BackgroundTask submission
+static IMP _BGTaskScheduler_submitTaskRequest_orig = NULL;
+
+static BOOL preSubmitTaskHook(id me, SEL selector, BGTaskRequest *req, NSError **perr) {
+  if([req isKindOfClass: [BGProcessingTaskRequest class]] &&
+     [req.identifier isEqualToString: @"app.covidshield.exposure-notification"]) {
```
jeberhardt, line 137: This is now the 4th time that "**app.covidshield.exposure-notification**" is in the codebase. It should be moved to a single spot. That may be tricky, considering it's in Typescript, Kotlin, Objective-C, and a info.plist file.

PR #918, opened 2020-07-27
Patch `async-storage` with `NSURLIsExcludedFromBackupKey`

jeberhardt:
## What

All files stored with `async-storage` were being stored in iCloud backups by default.

## Solution
Mark each file with the `NSURLIsExcludedFromBackupKey`. The easiest way to implement this was to patch-package the `async-storage` repo.

## Note
Affects iOS only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "fastlane:android:local": "bundle exec fastlane android local"
   },
   "dependencies": {
-    "@react-native-community/async-storage": "^1.9.0",
+    "@react-native-community/async-storage": "1.9.0",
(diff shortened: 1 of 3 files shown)


timarney:
cc: @obrien-j

henrytao-me:
Oh. So, we don't need the manifest check anymore?

jeberhardt:
> Oh. So, we don't need the manifest check anymore?

Yes, it's still there, just handled by the `else` clause.

obrien-j:
BVC-142 - iOS [app] User Data Backed Up To iCloud

This has been re-tested and marked as remediated by CCCS.

PR #921, opened 2020-07-27
Replaced "stop" with "slow" in FR

EliseKa:
Replaced "stop the spread" with "slow the spread" in French

PLEASE HOLD MERGING until we get the final green light.


EliseKa:
Closing. Direction changed, no longer need to do this

PR #922, opened 2020-07-27
Updated content on error screen iOS 14 beta users will see

smcmurtry:
Screen Shot 2020-07-27 at 3 24 47 PM


timarney:
Test on iOS simulator which doesn't have the Framework so should be good on this.

PR #946, opened 2020-07-29
Detox tests setup for Android

nickmly:
Test using:
`yarn pre:test:android` then `yarn test:android`

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -144,6 +144,8 @@ android {
         versionCode project.env.get("APP_VERSION_CODE") as Integer
         versionName project.env.get("APP_VERSION_NAME")
         resValue "string", "build_config_package", "app.covidshield"
+        testBuildType System.getProperty('testBuildType', 'debug')
+        testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
     }
     splits {
         abi {
@@ -179,6 +181,7 @@ android {
             minifyEnabled enableProguardInReleaseBuilds
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
             signingConfig signingConfigs.release
+            proguardFile "${rootProject.projectDir}/../node_modules/detox/android/detox/proguard-rules-app.pro"
         }
     }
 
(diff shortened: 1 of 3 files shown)


timarney:
Okay so this worked but ...

I had to download a new AVD I didn't have the Pixel 3 API 29 installed.

The test is also failing - which is separate from this PR.

@smcmurtry please test and see if you get similar results.  If it works - we can merge as this is to setup Android vs the failing test.

For both iOS and Android we should add a Detox testing page - linked from the Readme (not in the Readme) about how to set it up.  For now until we get further I will suggest adding a placeholder page to link to so it's not forgotten.

 
 

==================================================


detox[64184] INFO:  Example: has landing screen [FAIL]

 FAIL  e2e/firstTest.e2e.js (223.869 s)
  Example
    ✕ has landing screen (28962 ms)

```
Example › has landing screen

    Test Failed: 'at least 75 percent of the view's area is displayed to the user.' doesn't match the selected view.
    Expected: at least 75 percent of the view's area is displayed to the user.
         Got: null

    TIP: To print view hierarchy on failed actions/matches, use loglevel verbose and above.



detox[64183] ERROR: [cli.js] Error: Command failed: node_modules/.bin/jest --config e2e/config.json '--testNamePattern=^((?!:ios:).)*$' --maxWorkers 1 e2e
```

smcmurtry:
This worked for me! I changed the [.detoxrc.json](https://github.com/cds-snc/covid-alert-app/blob/master/.detoxrc.json#L18) to use the device I already had installed: `Pixel_3a_API_30_x86`. Question: should we move this value to the .env file, or should we all be using the same thing?

The test passed for me in 40 seconds. I had previously taken a snapshot on my emulator once it had reached a stable state, using [these instructions](https://github.com/wix/Detox/blob/master/docs/Introduction.AndroidDevEnv.md#emulator-quick-boot) from detox. I think that helps speed up the testing. I'll do a pr to add the documentation you suggested @timarney.

nickmly:
Ref: https://github.com/wix/Detox/issues/2234

smcmurtry:
> Question: should we move this value to the .env file, or should we all be using the same thing?

After discussion - we will all use the same emulator for now.

PR #949, opened 2020-07-29
Removing extraneous file.

jeberhardt:
This file was renamed to `ExposureConfigurationDefault.json`, but somehow this one, with the original name, stayed in the project.


smcmurtry:
I have verified that this file is now unused in the app. I think this should be approved and merged so that there is no confusion on what config to update in the future. But I will defer to @timarney here.

PR #950, opened 2020-07-29
Debug and release Info.plist files

nickmly:
Separates `Info.plist` into 2 different files for debug and release builds for Detox testing.


timarney:
@dsamojlenko can you have a look at this one?

We want to have a debug Info.plist so we don't end up committing a file back.  Want to avoid this https://github.com/cds-snc/covid-alert-app#ios-local-development

ios/CovidShield/Info-debug.plist
```diff
@@ -0,0 +1,70 @@
...
+<plist version="1.0">
+<dict>
+	<key>BGTaskSchedulerPermittedIdentifiers</key>
+	<array>
+		<string>app.covidshield.exposure-notification</string>
+	</array>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleDisplayName</key>
+	<string>COVID Alert</string>
```
jeberhardt, line 12: Since we are doing DEBUG, can we alter the display name? That way we know which app we are launching.

timarney: Good idea.  @nickmly ☝️

jeberhardt:
Are we going to apply the same tactic to Android?

dsamojlenko:
So, this just sets the default properties file to be `Info-debug.plist`. You would still need to remember to switch `Info.plist` and `Info-debug.plist` files when running release builds. Would be better if we had a way of automatically switching between the configurations.

Maybe a better approach would be using a build phase script like this: https://sveinhal.github.io/2016/02/23/disable-ats-in-debug-builds/

PR #951, opened 2020-07-29
Lint error on -> console log

timarney:
Add lint error for any console messages.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,7 @@ module.exports = {
     'import/no-named-as-default': 'off',
     '@shopify/images-no-direct-imports': 'off',
     '@shopify/jest/no-snapshots': 'off',
-    'no-console': 2,
+    'no-console': 'error',
     camelcase: 'off',
     'no-extend-native': 'off',
     'max-params': 'off',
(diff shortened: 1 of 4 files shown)


timarney:
cc: @dsamojlenko 

Screen Shot 2020-07-29 at 2 55 23 PM

jeberhardt:
This should also pop up a warning in IDEs such as WebStorm.

.gitignore
```diff
@@ -102,3 +102,5 @@ ios/assets
 .env.release
 
 artifacts/*
+
+coverage/*
```
jeberhardt, line 106: Technically, not a part of this PR.

timarney: removed

.eslintrc.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
     'import/no-named-as-default': 'off',
     '@shopify/images-no-direct-imports': 'off',
     '@shopify/jest/no-snapshots': 'off',
-    'no-console': 'off',
+    'no-console': 2,
```
jeberhardt, line 22: Perhaps using `error` would be more clear as to what this is doing?

timarney: using error now - good call

PR #957, opened 2020-07-30
Detox testing different home screens

smcmurtry:
I did some research into how we can reach different states in the app with detox. Here is what I tried:

- importing the async storage module into detox tests - this fails because ["there is no way detox can communicate with react native thread in runtime and change state"](https://stackoverflow.com/a/49528087/13161363)
- I started thinking through how we could load state from a json file that detox could write to. We would need to add a flag to the .env file that would tell the storage code to read state from a json file in the source code. But I don't think this would actually work, since at runtime the app is on the device and no longer has access to a file on your computer. 
- [The recommended way to mock](https://github.com/wix/detox/blob/master/docs/Guide.Mocking.md) in detox is to create a different version of each .js file you want to mock, with an .e2e.js extension, and then that gets used by the e2e tests. So we could create a different version of our app with a GUI way to reach different states, BUT....

We already have that - the test menu. So I think the best way to do this is to create a .env.e2e file that gets used in the e2e build, and then detox can open up the test menu and reach different screens the same way a human would. It also makes sense to have a dedicated .env.e2e file so we can use a build that connects to the staging server. We could probably get detox to do a real key claim this way if we want.

If you can think of another way to do this, please comment!


smcmurtry:
Android and iOS tests are now working. Sometimes iOS tests failing because of a `UIManager` warning popup that is covering the next button. Re-running the tests fixed the issue for me. All screenshots are being generated correctly for both android and ios.

PR #965, opened 2020-07-30
more unit tests backend service

pisajew:


src/services/BackendService/BackendService.spec.ts
```diff
@@ -117,6 +129,19 @@ describe('BackendService', () => {
           expect(covidshield.TemporaryExposureKey.create).toHaveBeenCalledWith(expect.objectContaining(value));
         });
     });
+
+    it('throws if random generator is not available', async () => {
```
henrytao-me, line 133: The unit test for `retrieveDiagnosisKeys` should be redone because it looks confusing to me. We should: 
- Stop at mocking `downloadDiagnosisKeysFile` and verify the `url` because that's what the scope of this function. 
- There are two cases we need to verify:
  + Use `00000` for last 14 days period. 
  + Passed period.

PR #969, opened 2020-07-30
Change exposureStatusType to an enum.

jeberhardt:
# Summary

This is a code optimization change. This changes the literal strings for exposureStatusType to an enum, so there are no magic strings. The functionality of the app should not be altered by this change. The goal is to make the code more maintainable.

# Test instructions

To test this change, run the app, and change the exposureStatus of the device from 'monitoring', to 'exposed' (by having it sit near a diagnosed phone for 15+ minutes. The exposureStatus should be changed and maintained between app launches.


timarney:
This looks great and moves us further maintainable code :)

Will make some time to test - @smcmurtry feel free to test as well.

timarney:
@jeberhardt let's get this conflict fixed up so we can merge this.

Issue #955, opened 2020-07-29
UI issue when viewing the app in landscape. The last checked label is not visible.


obrien-j:
![image](https://user-images.githubusercontent.com/17616596/88847181-acb3bb80-d1b4-11ea-9db8-c3b3be7e2c0b.png)

"Android UI suggestion would be for the bottom panel to be hidden when a user hit the back hardware key instead of closing the app."

source: https://twitter.com/ali_almunayer/status/1287566491924271104/photo/1

joshpc:
Design changes aside, it would be pretty straight forward to update `contentInset` (similar to what I proposed on https://github.com/cds-snc/covid-alert-app/pull/855) on these pages equal to the height of the menu to ensure that the entirety of the content is scrollable. It's not ideal since you can be playing with some pretty tiny screens and the menu will leave an unusable amount of space, but is a step in the right direction.

timarney:
Fixed in https://github.com/cds-snc/covid-alert-app/pull/971

PR #971, opened 2020-07-30
Fix scrolling menu

joshpc:
# Summary

When in landscape mode, some screens (specifically the No Exposure screen, as pointed out in https://github.com/cds-snc/covid-alert-app/issues/955) are clipped at the bottom by the overlay menu.

This is caused by the decision to have the `ScrollView` behind the `OverlayView`. This is a nice visual effect to let the content 'scroll behind' in a nice way. This could be implemented with `contentInsets` and without the weird gap, but a simpler fix here is to simply ensure that there is padding.

In the below screenshot, you see that the contentView (aka the content) only extends a portion of the screen, but the scroll view (blue) sits below only a portion of the menu. In a future PR that should probably be fixed / cleaned up but is very low priority.

Screen Shot 2020-07-30 at 6 23 35 PM

Additionally, this restores "proper scrolling" and bounciness to the `ScrollView` if and only if the content is longer than the scrollable area. This looks and feels better.

# Test instructions

The test here is to ensure that all screens are visible in both portrait and in landscape mode. There should be _zero_ content that cannot be reached.

Additionally, it should be tested that when in portrait, the screens are not scrollable UNLESS there is content below the menu.

Since this affects the `BaseHomeView` the following screens need to be viewed, and tested:
- No Exposure screens (No region, region covered, region not covered)
- Bluetooth Disabled
- Exposure Notifications disabled
- Framework Unavailable
- Network Disabled
- Exposed
- Diagnosed / Share
- Unknown Problem View

You should see behaviour like this on most pages.

**The view will be static. You cannot scroll this as there is not enough content!**
Screen Shot 2020-07-30 at 6 36 31 PM

**In landscape, you can scroll and it bounces.**
Please note that the colour changes in the menu was caused by the gif capture!
bounce

# Help requested

I was able to test this properly on iOS (simulator + iPhone X) but not on android (only emulator, no device.) Although emulator seemed fine.

As mentioned in the issue and linked issues, on very tiny screens regardless of if the content is scrollable or not, it's still difficult to see things. The menu is too fat!

# Reviewer checklist

- [ ] Is it accessible?
- [ ] Have you tested it?


smcmurtry:
Reviewing this PR I am noticing that we don't have consistent bottom padding on our home screen views:

`marginBottom="xl"` on last element in `ExposureView`:
Screen Shot 2020-08-19 at 11 03 58 AM

`marginBottom="m"` on last element in `NoExposureUncoveredRegionView`:
Screen Shot 2020-08-19 at 11 04 28 AM

This is not actually a problem specific to this PR though - I can fix this in a follow up PR.

timarney:
@smcmurtry let's fix in a follow up PR

PR #984, opened 2020-08-01
Replace prime marks with proper apostrophe marks

marktron:
# Summary

A few words in the English and French localizations use prime marks instead of apostrophes (e.g. `you're` instead of `you‘re`). This minor PR makes the copy consistently use apostrophes.

# Test instructions

These are just copy changes, so the normal build process should pull in the updated copy.

# Help requested

I don't actually know the purpose of `src/locale/translations/index.js`, so I didn’t update that file. I assumed it is generated during the build process, but if that’s not the case, it will need to be updated as well since it contains the same errors as `en.json` and `fr.json`.

# Reviewer checklist

- [ ] Is it accessible?
- [ ] Have you tested it?

### DIFF
--- a/src/locale/translations/en.json
+++ b/src/locale/translations/en.json
@@ -1,7 +1,7 @@
 {
   "Home": {
     "NoExposureDetected": {
-      "AllSetTitle": "You're all set",
+      "AllSetTitle": "You’re all set",
       "RegionCovered": {
         "Title": "No exposure detected",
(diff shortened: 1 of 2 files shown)


timarney:
> index.js ...  I assumed it is generated during the build process

That is correct

timarney:
@smcmurtry yes you can make a followup PR

PR #1004, opened 2020-08-05
use in-ecosystem URL for reachability, instead of leaking info to Google

scoates:
# Summary

The app leaks IP addresses to Google. This fixes it. Explanation in [issue](https://github.com/cds-snc/covid-alert-app/issues/1003).

# Test instructions

You'll need [this PR](https://github.com/cds-snc/covid-alert-server/pull/241) deployed before this can really be tested. Once ready, analyze traffic to be sure that no data is flowing to Google, and instead is staying local on `canada.ca`

# Reviewer checklist

This is a suggested checklist of questions reviewers might ask during their review:

- [ ] Does this meet a user need?
- [ ] Is it accessible?
- [ ] Is it translated between both offical languages?
- [ ] Is the code maintainable?
- [ ] Have you tested it?
- [ ] Are there automated tests?
- [ ] Does this cause automated test coverage to drop?
- [ ] Does this break existing functionality?
- [ ] Should this be split into smaller PRs to decrease change risk?
- [ ] Does this change the privacy policy?
- [ ] Does this introduce any security concerns?
- [ ] Does this significantly alter performance?
- [ ] What is the risk level of using added dependencies?
- [ ] Should any documentation be updated as a result of this? (i.e. README setup, etc.)

Issue #1003, opened 2020-08-05
URGENT: app contacts google.com and discloses IP addresses of all users


scoates:
# Summary

When the app runs (on iOS at least), it periodically requests `https://clients3.google.com/generate_204`. This is because the app uses [react-native-netinfo](https://github.com/react-native-community/react-native-netinfo), which, by default, exhibits this behaviour.

*Note:* I didn't report this as a security issue because it's not exploitable beyond what Google could already exploit. It just needs to be fixed ASAP.

**I urge you to treat this as an urgent security problem that must be fixed ASAP. You are currently leaking PII to Google for all users.**

See fixes below!

# Steps to reproduce

Running the app at all produces this behaviour, but in order to actually monitor it, you'll need to set up [mitmproxy](https://mitmproxy.org/) or otherwise watch what e.g. the iOS Simulator is doing via packet capture/Little Snitch, etc.

![](https://files.scoat.es/qbxC9S4q.png)

# Environment

I demonstrated the problem on an iOS device, but from what I can tell, this happens to all apps on all platforms that use react-native-netinfo (without changing the defaults).

# What's happening?

When the app invokes NetInfo [1](https://github.com/cds-snc/covid-alert-app/blob/336da088bf48cfaa3f65c309a2c53b6e609fbc5e/src/screens/home/HomeScreen.tsx#L2) [2](https://github.com/cds-snc/covid-alert-app/blob/aab62fd02542e19cf72402d105fc911733ee6aef/src/screens/home/views/OverlayView.tsx#L3), it uses an internal [Reachability Configuration](https://github.com/react-native-community/react-native-netinfo#netinfoconfiguration) to determine if the network is present and if it's behind a captive portal.

Unfortunately, the default URL that is used for this is on google.com.

# Why is this bad?

This is convenient for most React Native developers, I'm sure, but I'm also sure you'll admit that sending Google the IP addresses of all users, frequently, is bad form for an app that is specifically designed to track its users (and keep them safe while doing so). It's especially disturbing that such an app comes from our government.

## But really, is it so bad?

In a word? **Yes.**

Canada's Privacy Commissioner [says](https://www.priv.gc.ca/en/privacy-topics/privacy-laws-in-canada/the-personal-information-protection-and-electronic-documents-act-pipeda/pipeda-compliance-help/pipeda-interpretation-bulletins/interpretations_02/#fn50-rf):

[...]

PR #1008, opened 2020-08-06
[updated] use in-ecosystem URL for reachability, instead of leaking info to Google 

nickmly:
Copied from #1004. This PR fixes a CI issue that was occurring with #1004.

# Summary

The app leaks IP addresses to Google. This fixes it. Explanation in [issue](https://github.com/cds-snc/covid-alert-app/issues/1003).

# Test instructions

You'll need [this PR](https://github.com/cds-snc/covid-alert-server/pull/241) deployed before this can really be tested. Once ready, analyze traffic to be sure that no data is flowing to Google, and instead is staying local on `canada.ca`

# Reviewer checklist

- [ ] Is it accessible?
- [ ] Have you tested it?


dsamojlenko:
This is fine. I'll note that the reachability url does not have an effect on the app state. ie, replacing the 'presence' url with any dummy url would cause the reachability request not to fire at all, but our `network.isConnected` check would not fail.

scoates:
FWIW, I think that not firing the request (at least not so often) is probably even better. My main concern was the third party, though.

dsamojlenko:
Agreed. For reference, here's what network traffic at startup looks like with default config:
image

This is with our custom url enabled:
image

And here it is if we just provide an empty string for the reachabilityURL:
image

Code:
```
NetInfo.configure({
  reachabilityUrl: '',
});
```

Personally, I think fewer network calls is better.

scoates:
I'm on board if there's no other downside. Captive portals should still fail because of the https here, so the risk of injection is moot (or at least pushed down to the device's trust authority). Reduces load on the server, too.

scoates:
Thanks!

Is there somewhere I can read about how releases work after this? If not, is there a short version you can share?

PR #1010, opened 2020-08-06
Pass in full config to the en framework

smcmurtry:
This is to test backwards compatibility of the new config with the old app.


jeberhardt:
This is just for testing purposes.

PR #1019, opened 2020-08-06
Remove text about location permissions

voronaam:
Remove Android only text that explains to user a perculiar situation with the location permissions, which is no longer true.

The application does NOT request the `ACCESS_COARSE_LOCATION` permission anymore.

This is in line with the API documentation: https://developers.google.com/android/exposure-notifications/exposure-notifications-api

Hence the block no longer applies.

# Reviewer checklist

- [ ] Is it accessible?
- [ ] Have you tested it?

### DIFF
--- a/src/screens/home/views/ExposureNotificationsDisabledView.tsx
+++ b/src/screens/home/views/ExposureNotificationsDisabledView.tsx
@@ -42,19 +42,6 @@ export const ExposureNotificationsDisabledView = ({isBottomSheetExpanded}: {isBo
           onPress={onPress}
         />
       </Box>
-      {Platform.OS === 'android' ? (
-        <Box marginBottom="xl">
-          <Text marginBottom="m" variant="bodySubTitle">
-            {i18n.translate('Home.EnDisabled.AndroidTitle2')}
-          </Text>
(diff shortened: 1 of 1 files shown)


timarney:
Adding some reference links here:

**Current Android**

> The Exposure Notification technology uses Bluetooth scanning to understand what devices are near one another. On all phones running Android 6.0 and above, to use Bluetooth scanning, the device location setting needs to be turned on ...


https://support.google.com/android/answer/9930236?hl=en

**Android 11**

>On Android 11, which will soon be released, users will be able to use Exposure Notification apps without turning on the device location setting. We’re making this update for Exposure Notifications only, given that ENS has been designed in such a way that neither the system nor the apps using it can infer device location through Bluetooth scanning, and apps that are allowed to use ENS are subject to additional policies that disallow automatic collection of location. All other apps and services will still be prohibited from performing Bluetooth scanning unless the device location setting is on. 


> https://blog.google/inside-google/company-announcements/update-exposure-notifications

timarney:
Closing this based on https://github.com/cds-snc/covid-alert-app/pull/1019#issuecomment-671064551

PR #1023, opened 2020-08-09
Add initialNumToRender to Carousel

nima-ahmadi:
# Summary

This PR sets `initialNumToRender` prop of the onboarding Carousel to `1`. This change will improve the performance and the feel of the transition from the language selection screen to onboarding by initially loading the first page and load the reset of the pages after the transition.

Before/after demo:

covid

# Test instructions

- Clear Data
- Go through the onboarding flow
- The transition from _language selection_ to _onboarding carousel_ should be smoother and everything else stay be the same.

# Reviewer checklist

- [ ] Is it accessible?
- [x] Have you tested it?


timarney:
Thanks @nima-ahmadi

PR #1043, opened 2020-08-13
Add e2e region based tests

nickmly:
# Summary

Adds e2e detox tests for the region based screens (AllSetView, NoCode, NoExposureView).
In order to do this, DemoMode menu items were added to force the NoExposureViews.

# Test instructions

1. Run `yarn pre:test:ios` or `yarn pre:test:android` 
2. Run `yarn test:ios` or `yarn test:android`
3. Observe all tests pass

# Help requested

> Things that you (the submitter) want reviewers to pay very close attention to when they review this.

> Are there any related issues or tangent features you consider out of scope for this issue that could be addressed in the future?

# Reviewer checklist

- [ ] Is it accessible?
- [ ] Have you tested it?
- [x] Are there automated tests?


src/testMode/DemoMode.tsx
```diff
@@ -25,6 +25,10 @@ const ScreenRadioSelector = () => {
   const {forceScreen, setForceScreen} = useStorage();
   const screenData = [
     {displayName: 'None', value: 'None'},
+    {displayName: 'Not Exposed', value: 'NoExposureView'},
+    {displayName: 'Not Exposed (Region covered)', value: 'NoExposureCoveredRegionView'},
+    {displayName: 'Not Exposed (Region not covered)', value: 'NoExposureUncoveredRegionView'},
+    {displayName: 'Not Exposed (Region not set)', value: 'NoExposureNoRegionView'},
```
smcmurtry, line 31: It doesn't look like you're actually using these in the detox tests, is that correct? I would prefer if we don't add these so we don't have to expand the functionality of the test menu more than is necessary.

e2e/province.e2e.js
```diff
@@ -64,3 +78,55 @@ describe('Test province flow', () => {
...
+    await changeRegion('AB');
+    await element(by.id('tapPromptCollapsed')).tap();
+    await expect(element(by.id('getCodeButton'))).toBeVisible();
+    await element(by.id('getCodeButton')).tap();
+    await device.takeScreenshot('noCodeAB');
+    await expect(element(by.id('noCodeHeader'))).toBeVisible();
+    await expect(element(by.id('noCodeCTA'))).toBeNotVisible();
+    await element(by.id('toolbarCloseButton')).tap();
+    await element(by.id('bottom-sheet-close')).tap();
+  });
```
smcmurtry, line 102: I think this test will pass even if AB and ON display the same content on the no code screen. Could you check for either different text, or that the button only exists on the ON screen?

nickmly: @smcmurtry I fixed up the Android tests and I am using `await expect(element(by.id('noCodeCTA'))).not.toBeVisible();` to determine if the CTA button is visible on the No Code screen, is this sufficient for testing?

smcmurtry: Yes that is great!

PR #1044, opened 2020-08-13
Adhoc build distribution through Diawi

dsamojlenko:
# Summary

Adds Fastlane config to enable distributing AdHoc builds through [Diawi](https://www.diawi.com/). Also updates Fastlane to latest.

To distribute an AdHoc build, there are two new yarn commands:
- `yarn deploy-ios:diawi`
- `yarn deploy-android:diawi`

Once the build is complete, Fastlane will output a url that you can visit with a device to install the app. You can also log into your Diawi account and retrieve a QR Code for easy sharing.

See the [Diawi Docs](https://dashboard.diawi.com/docs/) for information on when links expire, how to share links, etc.

## Environment
By default, this script loads environment variables from the `.env.local`. Version numbers are less important in Adhoc distribution, but is displayed to users when downloading.

Likewise, the bundle id is not as important in AdHoc builds, but it should usually be the primary (production) bundle id.

## iOS Device registration
For iOS builds, this script will also register devices to the AdHoc profile in the Developer portal and regenerate/download the profile with those devices included. 

To enable this, you should include a `devices.txt` file in the `fastlane` folder following the file format outlined here: http://devimages.apple.com/downloads/devices/Multiple-Upload-Samples.zip

Note that this is an optimistic action- it will only ever add devices to the developer portal, and never remove devices.

## Diawi API
You will need to sign up for a (free) Diawi account and generate an API token which should be added to the `fastlane/.env` file as `DIAWI_TOKEN=...`


timarney:
@dsamojlenko @smcmurtry let's look at this and test on Monday.

PR #1047, opened 2020-08-13
Prototyping logic to narrow the contagious window

smcmurtry:
This draft PR prototypes some logic we could use to narrow the window around when TEKs are uploaded. We would obviously also need a way of getting the date of symptom onset before this would work.


timarney:
@smcmurtry going to close this for now and we can re-open or start a fresh PR once we're ready to go down this road.

PR #1050, opened 2020-08-14
Add `ENStatusPaused` to switch statement.

### LINKED ISSUE #1049: iOS introduced new ENStatus
# Summary
iOS 13.6 introduced a new ENStatus not being handled by `[ExposureConfiguration getStatus]`.

The new status is: `ENStatusPaused`
https://developer.apple.com/documentation/exposurenotification/enstatus/enstatuspaused

# Environment
iOS 13.6+

jeberhardt:
# Summary

Adding `ENStatusPaused` to switch statement.
Adding `default` case to switch statement, to ensure a value is always returned. Default value returned is `unknown`.
Closes #1049 

I have been unable to test the `paused` status, since it is not known how to pause **Exposure Notification** on the phone.

# Reviewer checklist

- [ ] Is it accessible?
- [ ] Have you tested it?

### DIFF
--- a/ios/CovidShield/ExposureNotification.m
+++ b/ios/CovidShield/ExposureNotification.m
@@ -68,8 +68,6 @@ RCT_REMAP_METHOD(stop, stopWithResolver:(RCTPromiseResolveBlock)resolve rejecter
 RCT_REMAP_METHOD(getStatus, getStatusWithResolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
 {
   switch (self.enManager.exposureNotificationStatus) {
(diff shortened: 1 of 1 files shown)


ios/CovidShield/ExposureNotification.m
```diff
@@ -78,6 +76,12 @@ - (void)invalidate
       break;
     case ENStatusRestricted: resolve(@"restricted");
       break;
+    case ENStatusPaused: resolve(@"paused");
```
timarney, line 79: We'll test this next week to see what happens but we might want to return unknown for this case.  At least until we have better info on the paused status.

PR #1052, opened 2020-08-17
Split the NoCode screen into separate views to simplify the logic

smcmurtry:
This will make it easier to add more logic to this screen in the future as we bring on more provinces and territories. This PR is purely a refactor - there should be no difference in the app (except for some bottom padding).

### DIFF
--- a/src/screens/nocode/NoCode.tsx
+++ b/src/screens/nocode/NoCode.tsx
@@ -1,138 +1,32 @@
 import React, {useCallback} from 'react';
-import {ScrollView, StyleSheet, Linking} from 'react-native';
-import {Box, Text, TextMultiline, Toolbar, ButtonSingleLine} from 'components';
+import {ScrollView, StyleSheet} from 'react-native';
+import {Box, Toolbar} from 'components';
 import {useI18n} from 'locale';
 import {useNavigation} from '@react-navigation/native';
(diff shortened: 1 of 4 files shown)


src/screens/nocode/NoCode.tsx
```diff
@@ -1,140 +1,34 @@
...
-    </Box>
-  );
-};
-
-export const NoCodeScreen = () => {
-  const i18n = useI18n();
-  const navigation = useNavigation();
-  const close = useCallback(() => navigation.goBack(), [navigation]);
-  const onChooseRegion = useCallback(() => navigation.navigate('RegionSelect'), [navigation]);
+import {NoRegionView} from './views/NoRegionView';
```
timarney, line 10: These should be at the top of the file

timarney:
Failing Detox test

Test one time code flow
    ✕ can get to the one time code screen (7341 ms)

  ● Test one time code flow › can get to the one time code screen

timarney:
Looks like we're missing the external link icon

src/screens/nocode/NoCode.tsx
smcmurtry: They are at the top - maybe it's all the deleted stuff in the diff making it look like they're farther down?

PR #1077, opened 2020-08-24
Pull active regions from .env

timarney:
This update moves the active regions array out to the .env file so we can push builds with new regions without needing to modify the code.

In the future we'll move this to a server call.


timarney:
@smcmurtry - looks like I based off the wrong branch.

PR #1087, opened 2020-08-27
Feature/add datepicker page

smcmurtry:
# Summary | Résumé

Will keep this a draft until #1085 is merged.

Added the "symptom onset screen". This includes the picker, which displays as a dropdown on Android and a picker wheel on iOS. On iOS the wheel is always open on the screen by default, which looks a bit weird. I am not addressing this yet but I did include a `ModalWrapper` component that could be used in the future to display the wheel in a modal. To format the dates that make up the options in this picker, I am using the `toLocaleString` function. This works out of the box on iOS but the version of javascript that is used for the Android build does not support this, so I needed to use the `date-time-format-timezone` and `intl` libraries there.

Revert this commit locally: https://github.com/cds-snc/covid-alert-app/commit/0b9ec61dfca5679528085d7db95b1b2fb97685c4. That will cause the "Enter your one-time key" button to link to the new screen. Note: the "Share exposures" button on this screen is not hooked up to anything yet.


smcmurtry:
Closing in favour of https://github.com/cds-snc/covid-alert-app/pull/1088 since this branch has a bunch of unneeded commits.

PR #1088, opened 2020-08-27
Add datepicker page (unlinked for now)

smcmurtry:
# Summary | Résumé

Added the "symptom onset screen". The screen is not linked to anywhere in the app, so this will change will not be visible in master for now. This includes the picker, which displays as a dropdown on Android and a picker wheel on iOS. On iOS the wheel is always open on the screen by default, which looks a bit weird. I am not addressing this yet but I did include a `ModalWrapper` component that could be used in the future to display the wheel in a modal. To format the dates that make up the options in this picker, I am using the `toLocaleString` function. This works out of the box on iOS but the version of javascript that is used for the Android build does not support this, so I needed to use the `date-time-format-timezone` and `intl` libraries there.

Android:
Screen Shot 2020-08-27 at 11 25 22 AM
Screen Shot 2020-08-27 at 11 25 34 AM

iOS:
Screen Shot 2020-08-27 at 1 30 30 PM

Revert [this commit](https://github.com/cds-snc/covid-alert-app/commit/922e83fd68550d864ec1a003e581c67c7ca8af58) locally. That will cause the "Enter your one-time key" button to link to the new screen. Note: the "Share exposures" button will only work if you are already in the diagnosed state.


smcmurtry:
Update - did some more work, put the iOS picker wheel in a modal. Styling still needs some work:
Screen Shot 2020-08-29 at 9 15 26 PM

timarney (dismissed):

timarney:
@smcmurtry we can merge after the conflicts are fixed up.

PR #1094, opened 2020-08-30
Traceback duration logic

smcmurtry:
# Summary | Résumé

I cherry-picked in the commits from #1047. I changed the logic to use the `contagiousDateInfo` object that will be passed in once #1088 is merged. The logic that determines when the user is contagious based on their test date needs looking at. I'm not sure how many days before the test date we are considering the person to be contagious (this has been set at 11 for now for no particular reason).

It may be easier to test this after #1088 is in. To test this PR by itself, you could manually change the symptom onset date in the source code, do the TEK upload flow through the app, and check that only TEKs from the contagious period are uploaded to the server.

There are a few TODOs that need resolving at some point.

### DIFF
--- a/src/services/BackendService/BackendService.ts
+++ b/src/services/BackendService/BackendService.ts
@@ -97,25 +97,35 @@ export class BackendService implements BackendInterface {
     };
   }
 
-  async reportDiagnosisKeys(keyPair: SubmissionKeySet, _exposureKeys: TemporaryExposureKey[]) {
+  filterTEKs = (key: TemporaryExposureKey) => {
+
+    /*
+      This function filters out TEKs that were generated outside the window
+      when the user was contagious.
+
+      rollingStartIntervalNumber = A number describing when a key starts. It is equal to startTimeOfKeySinceEpochInSecs / (60 * 10).
+
+      rollingPeriod = A number describing how long a key is valid. It is expressed in increments of 10 minutes (e.g. 144 for 24 hours).
+
+      source: https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/TemporaryExposureKey
+    */
     const symptomOnsetHoursSinceEpoch = hoursSinceEpoch(SYMPTOM_ONSET_DATE);
     const contagiousStartHoursSinceEpoch = symptomOnsetHoursSinceEpoch - CONTAGIOUS_DAYS_BEFORE_SYMPTOM_ONSET * 24;
 
+    const rollingEndIntervalNumber = key.rollingStartIntervalNumber + key.rollingPeriod;
+    const rollingEndIntervalHoursSinceEpoch = rollingEndIntervalNumber / 6;
+    if (rollingEndIntervalHoursSinceEpoch < contagiousStartHoursSinceEpoch) {
+      // the TEK is outside the contagious period
+      return false;
+    }
+    return true;
+  };
+
+  async reportDiagnosisKeys(keyPair: SubmissionKeySet, _exposureKeys: TemporaryExposureKey[]) {
     // Ref https://github.com/CovidShield/mobile/issues/192
     const filteredExposureKeys = Object.values(
       _exposureKeys
-        .filter(key => {
-          // rollingStartIntervalNumber = A number describing when a key starts. It is equal to startTimeOfKeySinceEpochInSecs / (60 * 10).
-          // rollingPeriod = A number describing how long a key is valid. It is expressed in increments of 10 minutes (e.g. 144 for 24 hours).
-          // source: https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/TemporaryExposureKey
-          const rollingEndIntervalNumber = key.rollingStartIntervalNumber + key.rollingPeriod;
-          const rollingEndIntervalHoursSinceEpoch = rollingEndIntervalNumber / 6;
-          if (rollingEndIntervalHoursSinceEpoch < contagiousStartHoursSinceEpoch) {
-            // the TEK is outside the contagious period
-            return false;
-          }
-          return true;
-        })
+        .filter(this.filterTEKs)
         .sort((first, second) => second.rollingStartIntervalNumber - first.rollingStartIntervalNumber),
     );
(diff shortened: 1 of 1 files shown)


src/services/BackendService/BackendService.ts
```diff
@@ -23,6 +23,16 @@ const TRANSMISSION_RISK_LEVEL = 1;
 // See https://github.com/cds-snc/covid-shield-server/pull/176
 const LAST_14_DAYS_PERIOD = '00000';
 
+const CONTAGIOUS_DAYS_BEFORE_SYMPTOM_ONSET = 2;
```
jeberhardt, line 26: These are the kind of numbers that should be included in **ExposureConfiguration.json**.

src/services/BackendService/BackendService.ts
```diff
@@ -23,6 +23,16 @@ const TRANSMISSION_RISK_LEVEL = 1;
...
+// todo: fix this - not sure this should be 11 days
+const CONTAGIOUS_DAYS_BEFORE_TEST_DATE = 11;
```
jeberhardt, line 28: These are the kind of numbers that should be included in **ExposureConfiguration.json**.

src/services/BackendService/BackendService.ts
```diff
@@ -92,10 +102,47 @@ export class BackendService implements BackendInterface {
...
+      // startTimeOfKeySinceEpochInSecs / (60 * 10).
+      // rollingPeriod = A number describing how long a key is valid. It is expressed in
+      // increments of 10 minutes (e.g. 144 for 24 hours).
+      // source: https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/TemporaryExposureKey
+      if (!contagiousDateInfo || contagiousDateInfo.dateType === 'noDate' || !contagiousDateInfo.dateString) {
+        return true;
+      }
+      // the following is a bit hacky - maybe we should pass the date as a Date, not a string
+      const dateParts = contagiousDateInfo.dateString.split('-');
+      const providedDate = new Date(Number(dateParts[0]), Number(dateParts[1]) - 1, Number(dateParts[2]));
```
jeberhardt, line 118: No need to use 'split' here; passing in a string to `new Date()` or `Date.parse()` should return a Date object.

src/services/BackendService/BackendService.ts
```diff
@@ -23,6 +23,16 @@ const TRANSMISSION_RISK_LEVEL = 1;
 // See https://github.com/cds-snc/covid-shield-server/pull/176
 const LAST_14_DAYS_PERIOD = '00000';
 
+const CONTAGIOUS_DAYS_BEFORE_SYMPTOM_ONSET = 2;
+// todo: fix this - not sure this should be 11 days
+const CONTAGIOUS_DAYS_BEFORE_TEST_DATE = 11;
+
+// todo: import this from BaseTekUploadView when PR #1088 is merged
+interface ContagiousDateInfo {
+  dateType: 'symptomOnsetDate' | 'testDate' | 'noDate';
```
jeberhardt, line 32: It would be great to use an Enum for `dateType`. That way there is no need to compare against a string later on in the code.

See `ExposureStatusType` in **ExposureNotificationService.ts** as an example.

src/services/BackendService/BackendService.ts
```diff
@@ -92,10 +102,47 @@ export class BackendService implements BackendInterface {
...
 
+  filterTEKs = (contagiousDateInfo?: ContagiousDateInfo) => {
+    return (key: TemporaryExposureKey) => {
+      // This function filters out TEKs that were generated before the user was contagious.
+      // rollingStartIntervalNumber = A number describing when a key starts. It is equal to
+      // startTimeOfKeySinceEpochInSecs / (60 * 10).
+      // rollingPeriod = A number describing how long a key is valid. It is expressed in
+      // increments of 10 minutes (e.g. 144 for 24 hours).
+      // source: https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/TemporaryExposureKey
+      if (!contagiousDateInfo || contagiousDateInfo.dateType === 'noDate' || !contagiousDateInfo.dateString) {
```
jeberhardt, line 113: This is where using an Enum for `dateType` would be useful.

src/services/BackendService/BackendService.ts
```diff
@@ -92,10 +102,47 @@ export class BackendService implements BackendInterface {
...
+      const providedDateHoursSinceEpoch = hoursSinceEpoch(providedDate);
+      let contagiousStartHoursSinceEpoch;
+      if (contagiousDateInfo.dateType === 'symptomOnsetDate') {
+        contagiousStartHoursSinceEpoch = providedDateHoursSinceEpoch - CONTAGIOUS_DAYS_BEFORE_SYMPTOM_ONSET * 24;
+      } else {
+        contagiousStartHoursSinceEpoch = providedDateHoursSinceEpoch - CONTAGIOUS_DAYS_BEFORE_TEST_DATE * 24;
+      }
+
+      const rollingEndIntervalNumber = key.rollingStartIntervalNumber + key.rollingPeriod;
+      const rollingEndIntervalHoursSinceEpoch = rollingEndIntervalNumber / 6;
```
jeberhardt, line 128: What is 6?

jeberhardt (changes requested):
See comments in the files.

src/services/BackendService/BackendService.ts
smcmurtry: `rollingEndIntervalNumber` is the number of 10 minute increments since epoch (see my comment [here](https://github.com/cds-snc/covid-alert-app/pull/1094/files/f9aae64e7df2e670487b3e6c6147231730878c57#diff-4faa9e63d84083d73a7e13a538ff797eR110)). So I am dividing by 6 to get that in hours.

jeberhardt: There should be a comment in the code to indicate that, or if the 6 is used somewhere else, then defined in a variable. Otherwise it's just a magic number.

smcmurtry:
Merged into #1109

PR #1108, opened 2020-09-08
Use .env for region.json

timarney:
This PR adds a new .env variable to `REGION_JSON_URL`

If set this URL will be used to fetch the region.json file

```
REGION_JSON_URL=https://raw.githubusercontent.com/cds-snc/covid-alert-app/feature/region-json-env/src/locale/translations/region.json
```

This will allow us to use a custom region.json for demos etc... i.e. when new provinces onboard.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,4 +9,4 @@ HMAC_KEY=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 MCC_CODE=302
 TEST_MODE=false
 MOCK_SERVER=false
-REGION_JSON_URL=
+REGION_JSON_URL=http://example.com/region.json # optional value for development mode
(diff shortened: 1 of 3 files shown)


.env.example
```diff
@@ -7,6 +7,6 @@ SUBMIT_URL=https://submission.covidshield.app
 RETRIEVE_URL=https://retrieval.covidshield.app
 HMAC_KEY=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 MCC_CODE=302
-
 TEST_MODE=false
 MOCK_SERVER=false
+REGION_JSON_URL=
```
jeberhardt, line 12: Should this value have an example URL?

timarney: Was debating this with myself.  Added now.

jeberhardt (approved):

PR #1119, opened 2020-09-15
Revising NB-specific content

katewilhelm:
Revising EN and FR content for NB based on today's meeting. Note that the URL on the exposed screen (both English and French) is still not live, so they're throwing a 404 error. Hopefully that's ok to update the URL now before the page is live.

# Summary | Résumé

> Updating NB-specific content and links in both English and French. 

---
To test, compare with mock-ups in figma ("WIP NB" page), check for type-os or weird spaces please.


smcmurtry (approved):
Looks good! We have a slight issue with spacing on the exposed screen but I will fix that in a follow up PR.

Screen Shot 2020-09-15 at 2 24 52 PM

PR #1132, opened 2020-09-22
Handle separate menu vs exposure screen links

dsamojlenko:
# Summary | Résumé

Handles rendering different "What to do if you're exposed" links for the Menu vs. Exposed screen. [See requirement from ON](https://trello.com/c/2i2OhVfR/551-set-up-app-to-have-different-links-for-menu-and-exposure-screens)

If provided, the Menu link will render the link from `RegionContent.Home.[PROVINCE].ExposedHelpLink`, otherwise it will fall back to either the provincial default at `RegionContent.ExposureView.[Active|Inactive].[PROVINCE].URL` or the national default at `RegionContent.ExposureView.Inactive.CA.URL`.

To handle this, added a `getExposedHelpMenuURL` method to `RegionLogic.ts` which will provide the custom URL if specified, or fallback to the default `getExposedHelpURL` method, which will return either the province-specific url, or the national one.

Since the region.json server config is pulled on every app start, you will need to add REGION_JSON_URL to your .env file. You can point it at the region.json in this PR:

```
REGION_JSON_URL=https://raw.githubusercontent.com/cds-snc/covid-alert-app/f5b4480f19f438b961a9be1fb1e4dda35b8fae5a/src/locale/translations/region.json
```

Since the new Menu urls are redirects, testing this can be a bit tricky - you'll need to watch closely for the redirect, or add some console.logs to see what's happening in debug mode.


smcmurtry (approved):
Nicely done!

PR #1135, opened 2020-09-24
Test push notifications

timarney:
Updating unit tests for sending reminders + exposed

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -180,6 +180,20 @@ export class ExposureNotificationService {
       });
   }
 
+  public pastMinimumReminderInterval(exposureStatus: any) {
+    if (exposureStatus.needsSubmission) {
+      return true;
+    }
+
+    if (!exposureStatus.uploadReminderLastSentAt) {
+      return true;
+    }
+
+    const lastSent = new Date(exposureStatus.uploadReminderLastSentAt);
+    console.log('lastSent', lastSent, new Date());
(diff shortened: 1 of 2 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -180,6 +180,23 @@ export class ExposureNotificationService {
...
   }
 
+  public isReminderNeeded(exposureStatus: ExposureStatus) {
+    if (exposureStatus.type !== ExposureStatusType.Diagnosed) {
+      return false;
+    }
+
+    if (exposureStatus.needsSubmission) {
+      return true;
+    }
```
smcmurtry, line 190: Comparing this to the old logic, I think it should be:
```
if (!exposureStatus.needsSubmission) {
  return false;
}
```

smcmurtry (changes requested):
Tests look good! Just one bit of logic I noticed needs fixing.

src/services/ExposureNotificationService/ExposureNotificationService.ts
timarney:
```
if (
      exposureStatus.type === ExposureStatusType.Diagnosed &&
      exposureStatus.needsSubmission &&
      (!exposureStatus.uploadReminderLastSentAt ||
        minutesBetween(new Date(exposureStatus.uploadReminderLastSentAt), new Date()) >
          MINIMUM_REMINDER_INTERVAL_MINUTES)
    ) {
```

smcmurtry (approved):
Looks good!

PR #1142, opened 2020-09-25
Mocking Platform.OS in our tests

smcmurtry:
# Summary | Résumé

This PR demonstrates how to mock `Platform.OS` in our jest tests so that it returns something other than `'ios'`. I also updated the tests here so they will work on android. The next step will be to figure out how to run all the tests on both android and ios in an automated way. We could do this by parameterizing each test but there is probably an easier way to do that.


timarney:
@smcmurtry let's chat more about this.

Thinking it would be nice to switch back and forth for some of the tests i.e. run as iOS , run as Android.

smcmurtry:
Well it looks like we can't add custom command line args to jest. So I just added the os as an extra parameter to some of the exposure tests so that they will automatically run for both android and ios. This is probably a better solution anyway since these tests will now be part of CI.

timarney (approved):

PR #1144, opened 2020-09-25
Fixing the return value of lastCheckedPeriod.

jeberhardt:
# Summary | Résumé

While refactoring some of the code, the incorrect value was returned from the `getKeys` function.


smcmurtry (approved):
Looks good - hooray for type checking!

PR #1152, opened 2020-09-29
Adds more realistic TEK to tests

maxneuvians:
Test improvement in that it adds more realistic TEKs to the tests

### DIFF
--- a/src/services/BackendService/BackendService.spec.ts
+++ b/src/services/BackendService/BackendService.spec.ts
@@ -1,4 +1,5 @@
 import nacl from 'tweetnacl';
+import crypto from 'crypto';
 import * as DateFns from 'shared/date-fns';
 import {blobFetch} from 'shared/fetch';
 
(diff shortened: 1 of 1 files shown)


src/services/BackendService/BackendService.spec.ts
```diff
@@ -86,12 +87,14 @@ expect.extend({
...
+  const today = (Math.floor(Date.now() / interval) * interval) / 1000;
   for (let i = 0; i < numberOfKeys; i++) {
     keys.push({
-      keyData: '',
-      rollingPeriod: i,
-      rollingStartIntervalNumber: i,
-      transmissionRiskLevel: 0,
+      keyData: crypto.randomBytes(16).toString('hex'),
+      rollingPeriod: 144,
+      rollingStartIntervalNumber: Math.floor((today - i * 86400) / (60 * 10)),
```
smcmurtry, line 96: Could you use `24 * 60 * 60` instead of `86400`? Not a huge deal but this would make it clearer to me how the math works.

smcmurtry (approved):
Looks good @maxneuvians! One minor comment.

maxneuvians:
Now that the main branch is good I will close and re-open with your suggestions @smcmurtry

PR #1157, opened 2020-09-29
Refactor `updateExposure` and `calculateNeedsSubmission`

jeberhardt:
# Summary | Résumé

An ongoing attempt to refactor the code to make it more manageable and easier to unit test. This pulls out the code to update the exposure status to a monitoring state.


timarney:
Have you completed an end to end test with these updates?

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -239,6 +239,54 @@ export class ExposureNotificationService {
...
+
+    const submissionLastCompletedAt = exposureStatus.submissionLastCompletedAt;
+    if (!submissionLastCompletedAt) {
+      return true;
+    }
+
+    const lastSubmittedDay = new Date(submissionLastCompletedAt);
+
+    // This has to be based on UTC timezone https://github.com/cds-snc/covid-shield-mobile/issues/676
+    return daysBetweenUTC(lastSubmittedDay, today) > 0;
```
timarney, line 258: Given we're looking to return a Boolean here - would be good to double check what happens when bad data is given to daysBetweenUTC to see what it returns.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -239,6 +239,54 @@ export class ExposureNotificationService {
     await this.recordKeySubmission();
   }
 
+  calculateNeedsSubmission(exposureStatus: ExposureStatus, today: Date): boolean {
+    if (exposureStatus.type !== ExposureStatusType.Diagnosed) return false;
```
timarney, line 243: Just noting the change here we used to have this inline.  Unclear if given the Observable the value could possibly be different vs the parameter (would need some testing).

```
```

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -239,6 +239,54 @@ export class ExposureNotificationService {
...
+    if (exposureStatus.type === ExposureStatusType.Monitoring) return exposureStatus;
+    switch (exposureStatus.type) {
+      case ExposureStatusType.Diagnosed:
+        // There is a case where using UTC and device timezone could mess up user experience. See `date-fn.spec.ts`
+        // Let's use device timezone for resetting exposureStatus for now
+        // Ref https://github.com/cds-snc/covid-shield-mobile/issues/676
+        if (daysBetween(today, new Date(exposureStatus.cycleEndsAt)) <= 0) {
+          return {type: ExposureStatusType.Monitoring, lastChecked: exposureStatus.lastChecked};
+        } else {
+          return Object.assign(exposureStatus, {
```
timarney, line 271: Could use the spread operator here as it's used for most of the codebase
```
{...exposureStatus, }
```

src/services/ExposureNotificationService/ExposureNotificationService.ts
timarney: Old

``` 
if (currentStatus.type === ExposureStatusType.Diagnosed) {
      const today = getCurrentDate();
      const cycleEndsAt = new Date(currentStatus.cycleEndsAt);
      // There is a case where using UTC and device timezone could mess up user experience. See `date-fn.spec.ts`
      // Let's use device timezone for resetting exposureStatus for now
      // Ref https://github.com/cds-snc/covid-shield-mobile/issues/676
      if (daysBetween(today, cycleEndsAt) <= 0) {
        this.exposureStatus.set({type: ExposureStatusType.Monitoring, lastChecked: currentStatus.lastChecked});
        return this.finalize();
      }
```

New including Master - looks like this would reach detectExposure when Diagnosed vs the old code wouldn't


``
 const updatedExposure = this.updateExposure(currentExposureStatus, today);
    if (updatedExposure !== currentExposureStatus) {
      this.exposureStatus.set(updatedExposure);
    }

    const {keysFileUrls, lastCheckedPeriod} = await this.getKeysFileUrls();

    try {
      captureMessage('lastCheckedPeriod', {lastCheckedPeriod});
      const summaries = await this.exposureNotification.detectExposure(exposureConfiguration, keysFileUrls);
```

**Comment by timarney** [comment]
@smcmurtry please review as well.

File: src/services/ExposureNotificationService/ExposureNotificationService.ts (hunk 1 above)
Comment (jeberhardt) [reply src/services/ExposureNotificationService/ExposureNotificationService.ts:265]: > Given we're looking to return a Boolean here - would be good to double check what happens when bad data is given to daysBetweenUTC to see what it returns.

I've added in some checks to ensure that the date values being used are actually dates.

**Comment by jeberhardt** [comment]
> Have you completed an end to end test with these updates?

Yes. Successful Android --> Android test over 20 minutes.

File: src/services/ExposureNotificationService/ExposureNotificationService.ts [hunk 4]
```diff
@@ -349,26 +415,12 @@ export class ExposureNotificationService {
...
-      const lastExposureAt = new Date(currentStatus.summary.lastExposureTimestamp || today.getTime());
-      if (daysBetween(lastExposureAt, today) >= EXPOSURE_NOTIFICATION_CYCLE) {
-        this.exposureStatus.set({type: ExposureStatusType.Monitoring, lastChecked: currentStatus.lastChecked});
-        return this.finalize();
-      }
+    const currentExposureStatus = this.exposureStatus.get();
+    const today = getCurrentDate();
+    const updatedExposure = this.updateExposure(currentExposureStatus, today);
+    if (updatedExposure !== currentExposureStatus) {
+      this.exposureStatus.set(updatedExposure);
```
smcmurtry, line 423: I think you should also have a `this.finalize()` call here to match the old code. That updates the `lastChecked` object in the status, and has some extra logic in case `lastChecked` on the current status is missing values.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -256,6 +256,62 @@ export class ExposureNotificationService {
...
+    const lastSubmittedDay = new Date(submissionLastCompletedAt);
+
+    if (isNaN(lastSubmittedDay.getTime())) return true;
+
+    // This has to be based on UTC timezone https://github.com/cds-snc/covid-shield-mobile/issues/676
+    return daysBetweenUTC(lastSubmittedDay, today) > 0;
+  }
+
+  public updateExposure(exposureStatus: ExposureStatus, today: Date): ExposureStatus {
+    if (exposureStatus.type === ExposureStatusType.Monitoring) return exposureStatus;
```
smcmurtry, line 286: This could just be another case statement for consistency

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -256,6 +256,62 @@ export class ExposureNotificationService {
...
+          return {type: ExposureStatusType.Monitoring, lastChecked: exposureStatus.lastChecked};
+        } else {
+          return {
+            ...exposureStatus,
+            needsSubmission: this.calculateNeedsSubmission(exposureStatus, today),
+          };
+        }
+      case ExposureStatusType.Exposed:
+        if (
+          daysBetween(new Date(exposureStatus.summary.lastExposureTimestamp || today.getTime()), today) >=
```
smcmurtry, line 302: I find this line a bit hard to read. Would you be okay with just defining the variable in the case block like the old code does?
```
const lastExposureAt = new Date(currentStatus.summary.lastExposureTimestamp || today.getTime());
```

That would require adding a lint exception but is worth it IMHO.

jeberhardt: It's a `const` that is only used once, so storing it in a variable isn't really necessary. Perhaps there's another way to make it easier to read.

src/services/ExposureNotificationService/ExposureNotificationService.ts
jeberhardt: This was added as an early return, but you're right, it doesn't fit the rest of the code.

jeberhardt:
Replaced with PR #1169

PR #1166, opened 2020-10-01
Simplified getKeysFileUrls, reduced # of calls to the framework

smcmurtry:
- removed generator, 
- made a simpler `getPeriodsSinceLastFetch` fn to replace `keysSinceLastFetch`
- reduced # of calls to the framework by calling `this.backendInterface.retrieveDiagnosisKeys` with `0` when the app is 1st installed, instead of 14 times

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -370,7 +370,7 @@ export class ExposureNotificationService {
         return this.finalize();
       }
     }
-    const {keysFileUrls, lastCheckedPeriod} = await this.getKeysFileUrls();
+    const {keysFileUrls, lastCheckedPeriod} = await this.getKeysFileUrls(currentStatus);
 
     try {
       captureMessage('lastCheckedPeriod', {lastCheckedPeriod});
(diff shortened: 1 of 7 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -411,20 +396,16 @@ export class ExposureNotificationService {
     return this.finalize();
   }
 
-  private async getKeysFileUrls() {
+  private async getKeysFileUrls(): Promise<any> {
     const currentStatus = this.exposureStatus.get();
```
jeberhardt, line 400: Can `currentStatus` be passed into this function? A tiny bit more functional programming that way.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -411,20 +396,16 @@ export class ExposureNotificationService {
...
     let lastCheckedPeriod = currentStatus.lastChecked?.period;
-    while (true) {
-      const {value, done} = await generator.next();
-      if (done) break;
-      if (!value) continue;
-      const {keysFileUrl, period} = value;
-      captureMessage('loop period', {period});
+    const periodsSinceLastFetch = this.getPeriodsSinceLastFetch(lastCheckedPeriod);
+    for (const period of periodsSinceLastFetch) {
+      const keysFileUrl = await this.backendInterface.retrieveDiagnosisKeys(period);
```
maxneuvians, line 405: It looks like the old code wrapped this in a try catch - the danger here might be that a server or network error goes unnoticed

smcmurtry: Good catch!

maxneuvians (approved):
Nice work!

PR #1169, opened 2020-10-02
Refactoring `updateExposure` and `calculateNeedsSubmission`

jeberhardt:
# Summary | Résumé

An ongoing attempt to refactor the code to make it more manageable and easier to unit test. This pulls out the code to update the exposure status to a monitoring state.

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.spec.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.spec.ts
@@ -774,24 +774,39 @@ describe('ExposureNotificationService', () => {
     const today = new OriginalDate();
     service.exposureStatus.set({
       type: ExposureStatusType.Diagnosed,
-      cycleStartsAt: today.getDate() - 15 * ONE_DAY,
-      cycleEndsAt: today.getDate() - ONE_DAY,
+      cycleStartsAt: today.valueOf() - 15 * ONE_DAY,
+      cycleEndsAt: today.valueOf() - ONE_DAY,
       needsSubmission: true,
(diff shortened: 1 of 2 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.spec.ts
```diff
@@ -749,6 +749,63 @@ describe('ExposureNotificationService', () => {
...
+      }),
+    });
+    expect(service.calculateNeedsSubmission()).toStrictEqual(false);
+  });
+
+  it('calculateNeedsSubmission when diagnosed false', () => {
+    service.exposureStatus.set({
+      type: ExposureStatusType.Diagnosed,
+      cycleStartsAt: new OriginalDate('2020-05-18T04:10:00+0000').getDate(),
+      cycleEndsAt: 0,
```
smcmurtry, line 777: `cycleEndsAt` should be 14 days after `cycleStartsAt` to make this more realistic. You could also do it more like the other test you wrote to simulate a cycle that has just finished:
```
      cycleStartsAt: today.getDate() - 15 * ONE_DAY,
      cycleEndsAt: today.getDate() - 1 * ONE_DAY,
```

smcmurtry (changes requested):
Code looks good. Could you add one more test to cover the `exposureStatus.submissionLastCompletedAt` logic? i.e. that `calculateNeedsSubmission` returns false if the user is diagnosed but has uploaded data already that day.

smcmurtry (approved):
Looks good!!

PR #1170, opened 2020-10-02
Removed ability to change UUID from test menu, added copy to clipboard functionality

smcmurtry:
# Summary | Résumé

This PR removes the ability to change the Debug UUID and adds a copy UUID to clipboard button. It turns out we don't actually want people to change the UUID during a test run. If it gets changed, that means we don't have an easily searchable, complete log of all events coming from that device any more. So best to remove the feature and make it easier to copy the existing UUID so it can be sent out.

I tested the copy functionality on Android 9 and iOS 13.7.

Screen Shot 2020-10-02 at 3 30 21 PM

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -310,6 +310,8 @@ PODS:
     - React
   - RNCAsyncStorage (1.9.0):
     - React
+  - RNCClipboard (1.3.0):
+    - React-Core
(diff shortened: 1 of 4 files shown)


timarney:
@dsamojlenko thinking we can add this to the loggly setup for local testing (not in the main branch - debug only).

PR #1177, opened 2020-10-06
Fix typo in french

xel1045:
The word "vivants" should have an "s" at the end, because there is more than one "personnes" and "animaux".


timarney:
@EliseKa ☝️ can you review?

100terres (approved):

EliseKa:
No, there is no typo in the original version. "vivant" should not have an "s" as it is a present participle, not an adjective.

xel1045:
Oops, a present participle is indeed invariant. Sorry for that!

PR #1185, opened 2020-10-06
feat: use etag on region.json

maxneuvians:
# Summary | Résumé

The PR adds in the ability for the app to use Etags when polling for new content from the CDN, specifically the `region.json` file. 

It add a function that wraps `fetch` which will use Etags when is queries the URL. If a 200 is returned, then it saves any Etags sent and caches a local version of the content inside the app. If a 304 is returned, it retrieves the cached copy. It is URL agnostic, so it should work both for the `region.json` and any other URLs called.

There is a quirk in **iOS** that is of note, which is that react-native fetch uses the browser as a pass through on caching. This has the following side effect:

1. App calls browser to query server, server returns a 200 status with a body and an Etag, this is passed directly to app and cached in the browser according to server cache rules.

2. App calls browser to query server with stored Etag, server returns a 304 with no body, browser passes cached version back to app with the same status code of 200 and body.

This means we can't rely on `response.status` ever being a 304 on iOS, because the browser has already dealt with the caching. This was validate through two ways:

1. Using a man in the middle proxy I was able to observe the server returning a 304, yet the app received a 200 status code from the browser.

2. I set up express.js server, where on the first call it returned a 200 status code with a body. Then I changed the server to only reply with a 304 with an empty body. On the subsequent call the app showed the original 200 status code and the original body.

In **Android** there seems to be no caching in the browser. 200 and 304 status codes are passed transparently.

- This should also be implemented for `CA.json`.

# Reviewer checklist | Liste de vérification du réviseur

This is a suggested checklist of questions reviewers might ask during their review | Voici une suggestion de liste de vérification comprenant des questions que les réviseurs pourraient poser pendant leur examen :

- [ ] Does this meet a user need? | Est-ce que ça répond à un besoin utilisateur?
- [ ] Is it accessible? | Est-ce que c’est accessible?
- [ ] Is it translated between both offical languages? | Est-ce dans les deux langues officielles?
- [ ] Is the code maintainable? | Est-ce que le code peut être maintenu?
- [ ] Have you tested it? | L’avez-vous testé?
- [ ] Are there automated tests? | Y a-t-il des tests automatisés?
- [ ] Does this cause automated test coverage to drop? | Est-ce que ça entraîne une baisse de la quantité de code couvert par les tests automatisés?
- [ ] Does this break existing functionality? | Est-ce que ça brise une fonctionnalité existante?
- [ ] Should this be split into smaller PRs to decrease change risk? | Est-ce que ça devrait être divisé en de plus petites demandes de tirage (« pull requests ») afin de réduire le risque lié aux modifications?
- [ ] Does this change the privacy policy? | Est-ce que ça entraîne une modification de la politique de confidentialité?
- [ ] Does this introduce any security concerns? | Est-ce que ça introduit des préoccupations liées à la sécurité?
- [ ] Does this significantly alter performance? | Est-ce que ça modifie de façon importante la performance?
- [ ] What is the risk level of using added dependencies? | Quel est le degré de risque d’utiliser des dépendances ajoutées?
- [ ] Should any documentation be updated as a result of this? (i.e. README setup, etc.) | Faudra-t-il mettre à jour la documentation à la suite de ce changement (fichier README, etc.)?


timarney:
Closing we may revisit in the future

Issue #1130, opened 2020-09-22
App should not request location to be enabled on Android 11


ajayyy:
This requirement was removed on Android 11.

https://developers.google.com/android/exposure-notifications/release-notes#locationless_scanning

The generic `deviceSupportsLocationlessScanning()` can be used

timarney:
Noting this is on our radar and is in our backlog :)

Will update here in the future.

100terres:
@timarney 

I personally almost uninstalled the app because I did not understand why it was "deactivated". When I clicked on "Activer Alert COVID" it only explain that I need to active the Bluetooth nothing about the localization. There is a text below the CTA explaining the location need to be activated... But I think it is not clear enough !

![signal-2020-10-06-143416](https://user-images.githubusercontent.com/4420524/95246255-36e65480-07e2-11eb-9326-9710845bc7ee.png)
![signal-2020-10-06-143753](https://user-images.githubusercontent.com/4420524/95246254-364dbe00-07e2-11eb-99ac-f45ee42912f4.png)
![signal-2020-10-06-144206](https://user-images.githubusercontent.com/4420524/95246250-364dbe00-07e2-11eb-8773-200da51b4e15.png)

timarney:
@100terres @ajayyy it's on the priority list - we'll get to it as soon as we possibly can.

100terres:
Thank you ! @timarney :upside_down_face:

timarney:
FYI - This rolled out in the latest release 1.0.12.

PR #1187, opened 2020-10-07
Added code to enable `locationless scanning` for Android.

jeberhardt:
# Summary | Résumé

Android 11 introduces "locationless scanning" for supported devices. This allows the app to work without requiring "location services" to be enabled for the app or the phone. This simple check in the code uses an API that was introduced in 1.6 of the Exposure Notification framework.

This should be tested on both Android 11 and Android 10 or below.

- Launch the app and verify everything is working correctly.
- Minimize the app.
- Turn off Location Services.
- Return to the app: 
-- on Android 11, the app should still be functioning.
-- on Android 10 or below, the app should notify you that COVID Alert is Off, with instructions to turn it back on.

A full end-to-end test should also be performed, on Android 11, with Location Services turned off.

# Help requested | Aide requise

Please pay close attention to the behaviour on Android 11, as I do not yet have a phone that supports Android 11.

## This change is for Android only. It has no impact on iOS and there should be no need to test on iOS.

### DIFF
--- a/android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
+++ b/android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
@@ -7,8 +7,6 @@ import android.content.Context
 import android.content.Intent
 import android.content.IntentSender
 import android.location.LocationManager
-import android.os.Build.VERSION;
-import android.os.Build.VERSION_CODES;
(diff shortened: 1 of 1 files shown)


LazyEngineerToBe:
This would fix issue #1130

timarney (approved):

PR #1189, opened 2020-10-07
Enter code text updates

timarney:
Screen Shot 2020-10-07 at 12 54 58 PM


timarney:
En



Fr

smcmurtry (approved):
Looks good. Nice catch on the left alignment - that was a bug that we had not found since none of our titles were long enough to wrap.

PR #1190, opened 2020-10-09
PEI => Update region.json

timarney:


smcmurtry (approved):
Looks good - matches current prod server.

PR #1193, opened 2020-10-09
Feature/v2 adapters

smcmurtry:
# Summary | Résumé

This mocks 2 of the new V2 methods in the adapters: `provideDiagnosisKeys` and `getExposureWindows`. 
I left todo comments to replace these mocks after the native code is implemented. I figure this will let us move forward and implement new methods in `ExposureNotificationService` and write tests, so we can have the code in master but not switch it on until a later date.


smcmurtry:
Closing since we should keep this data mocking out of the app - it would be better done in the tests.

PR #1194, opened 2020-10-09
Add V2 filtering logic that detects exposures given an array of `ExposureWindow`s

smcmurtry:
# Summary | Résumé

This PR adds V2 filtering logic that detects exposures given an array of `ExposureWindow`s. I have added:
- `checkIfExposedV2()`: this takes an array of `ExposureWindow`s, bluetooth attenuation thresholds, and the 15 minute minimum and checks if the provided array contains an exposure
- `performExposureStatusUpdateV2()` is meant to be interchangeable with the existing function `performExposureStatusUpdate()`, but it uses the new V2 functionality

These functions are not called anywhere yet and should not be until we have implemented `getExposureWindows()` and `provideDiagnosisKeys()` in the native layer. I am testing `checkIfExposedV2()` w/ mock data.

This is one approach - an alternative would be to use the new `getDailySummaries()` function.

[Google V2 API docs](https://developers.google.com/android/exposure-notifications/exposure-notifications-api).


smcmurtry:
Posting some review comments from others:
- split V2 tests into a separate file
- pass args to `checkIfExposedV2` as an object for readability
- split the scan instance logic in the for loop into its own fn
- populate a new summary obj to save in the status, or come up w/ an alternative (may need to calc the last exposed timestamp`)
- sort exposure windows in order of most recent to help w/ the above

smcmurtry:
A few more todos:
- [ ] save a "fake" summary obj to the status in the V2 code, this will require:
- [ ] calculate `matchedKeyCount`
- [ ] calculate `lastExposureTimestamp`
- [ ] calculate `attenuationDurations`
- [ ] write test to check `matchedKeyCount` is correct
- [ ] write test to check `lastExposureTimestamp` is correct
- [ ] write tests for "multiple exposures in the same day" scenario

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -341,6 +344,120 @@ export class ExposureNotificationService {
...
+      .map(x => x.secondsSinceLastScan)
+      // with initial value to avoid when the array is empty
+      .reduce((partialSum, x) => partialSum + x, 0);
+    return totalSeconds;
+  };
+
+  public getAttenuationDurations = (scanInstances: ScanInstance[], attenuationDurationThresholds: number[]) => {
+    //  typicalAttenuation values are in positive DB units
+    const immediateScans = scanInstances.filter(x => x.typicalAttenuation <= attenuationDurationThresholds[0]);
+    const nearScans = scanInstances.filter(x => {
```
jeberhardt, line 358: Please change the variable "x" to something more meaningful.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -341,6 +344,120 @@ export class ExposureNotificationService {
...
+    attenuationDurationThresholds: number[];
+    minimumExposureDurationMinutes: number;
+  }): Promise<[boolean, ExposureSummary | undefined]> {
+    if (exposureWindows.length === 0) {
+      return [false, undefined];
+    }
+    const dailySummariesObj: {[key: string]: {attenuationDurations: number[]; matchedKeyCount: number}} = {};
+    exposureWindows
+      .filter(window => {
+        if (window.infectiousness === Infectiousness.None) {
```
jeberhardt, line 383: This `if` statement can be simplified, combined with the `return` statement.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -341,6 +344,120 @@ export class ExposureNotificationService {
...
+      .forEach(windowSummary => {
+        const dayString = windowSummary.day.toString();
+        dailySummariesObj[dayString].attenuationDurations[0] += windowSummary.attenuationDurations[0];
+        dailySummariesObj[dayString].attenuationDurations[1] += windowSummary.attenuationDurations[1];
+        dailySummariesObj[dayString].attenuationDurations[2] += windowSummary.attenuationDurations[2];
+        dailySummariesObj[dayString].matchedKeyCount += 1;
+      });
+
+    const dailySummaries = Object.entries(dailySummariesObj)
+      .map(x => {
```
jeberhardt, line 404: Please change the variable "x" to something more meaningful.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -341,6 +344,120 @@ export class ExposureNotificationService {
...
+          {
+            type: ExposureStatusType.Exposed,
+            summary: dailySummary,
+          },
+          lastCheckedPeriod,
+        );
+      }
+      return this.finalize({}, lastCheckedPeriod);
+    } catch (error) {
+      captureException('performExposureStatusUpdateV2', error);
```
jeberhardt, line 455: Should this `return false` to reject the promise?

jeberhardt (changes requested):

src/services/ExposureNotificationService/ExposureNotificationService.ts
smcmurtry: I find my version easier to read but I will go with your suggestion anyway 😉

src/services/ExposureNotificationService/ExposureNotificationService.ts
jeberhardt: Shouldn't it return a boolean value, rather than just void?

jeberhardt (approved):

PR #1197, opened 2020-10-13
Removing `BackgroundFetch.scheduleTask()` call on Android.

jeberhardt:
# Summary | Résumé

Removing the BackgroundFetch.scheduleTask() call on Android, since it is unnecessary and redundant. iOS relies on 2 different types of background tasks, whereas Android treats them both the same. BackgroundFetch.configure() not only configures the task, but also initiates it.

- Install the app on Android and get past onboarding.
- The background checks should happen every 4-8 hours.


jeberhardt:
[Dig into background task on Android](https://trello.com/c/Hvpc4W4C/596-dig-into-background-task-on-android)

timarney:
Testing looks good so far.

timarney (approved):

PR #1201, opened 2020-10-15
Add additional summary filter

timarney:


smcmurtry (approved):
Looks good! This will make the filtering extra safe.

PR #1202, opened 2020-10-15
Update region.json 

timarney:


smcmurtry (approved):
Looks good! I checked the diff and this matches what is in prod.

PR #1207, opened 2020-10-16
Remove the option to use AlarmManager for Android.

jeberhardt:
# Summary | Résumé

The `BackgroundFetch` package can use AlarmManager for Android. In early development testing, this was a convenient way of testing background tasks on Android, and getting them to execute more frequently. However, since the production code will always use JobScheduler to execute background tasks, it is recommended that developers do not bypass that code for the sake of accelerating the testing cycle.

It is still possible to run the AlarmManager version of the background task, but it should be done by explicitly setting `forceAlarmManager` to true, and not be tied to the .env variable `TEST_MODE`.

NOTE: This change affects Android only, and only in the development environment, not in production.

- Set `TEST_MODE` in the .env file to be **TRUE**.
- `yarn run-android`
- In Android Studio, place a breakpoint on line 116 of the decompiled class file **BGTask.class**.
- Build and run for Android in Debug mode.
- Upon hitting the breakpoint, observe that the code for JobScheduler is executed, and not the code for AlarmManager.


timarney (approved):

jeberhardt:
[Dig into background task on Android](https://trello.com/c/Hvpc4W4C/596-dig-into-background-task-on-android)

PR #1208, opened 2020-10-16
Proof of Concept - Sharing the app

dsamojlenko:
# Summary | Résumé

Quick proof-of-concept for app sharing.

Uses the [react-native Share API](https://reactnative.dev/docs/share) which passes content to the OS-level share sheet.

To test, run a debug build with the TestMenu enabled. You'll find a "Share this app" button in the menu.

The Share Sheet will look different depending on the apps installed on the device. ie, if the user has Twitter installed, it will show up in the list. If they don't have Facebook installed, it will not appear in the list, etc. Users can also customize their share sheets.

The various apps will treat the content differently. For example, for emails we can provide a Subject line, but this doesn't work on all email clients. Messages on iOS renders a preview of the app store listing. Twitter doesn't seem to render a social card, at least not in the preview - it may 'unfurl' in the tweet when sent. Not sure about FB.

Here's what the share sheet looks like on each os:

iOS | Android
----- | -----
image | image

### DIFF
--- a/src/screens/testScreen/TestScreen.tsx
+++ b/src/screens/testScreen/TestScreen.tsx
@@ -1,5 +1,5 @@
 import React, {useCallback, useState, useEffect} from 'react';
-import {TextInput, StyleSheet, ScrollView, Share} from 'react-native';
+import {TextInput, StyleSheet, ScrollView, Share, Platform} from 'react-native';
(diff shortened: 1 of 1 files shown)


timarney:
Closing this for now.

PR #1211, opened 2020-10-20
Feature/fastlane latest version

dsamojlenko:
# Summary | Résumé

Adds a convenience function to Fastlane to get the latest used Version Codes (Build #s) from the App stores. Makes for quicker lookup rather than logging into both stores and reviewing both bundle ids manually.

To use it, run `yarn get_latest_version` and after a few minutes, you should see something like this:

image


timarney:
download

timarney (approved):
This is great @dsamojlenko

PR #1229, opened 2020-11-03
Feature/v2 en native layer

jeberhardt:
# Summary | Résumé

Adding the interface for react-native to make these API calls in Android and iOS.

Testing requires  PR #1194 that implements the `react-native` code. See that PR for testing instructions.

### DIFF
--- a/android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
+++ b/android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
@@ -206,6 +206,7 @@ class ExposureNotificationModule(context: ReactApplicationContext) : ReactContex
                 val files = diagnosisKeysURLs.parse(String::class.java).map { File(it) }
                 val diagnosisKeyFileProvider = DiagnosisKeyFileProvider(files)
                 exposureNotificationClient.provideDiagnosisKeys(diagnosisKeyFileProvider)
+                promise.resolve(null)
             } catch (exception: Exception) {
                 promise.reject(exception)
             }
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/app/covidshield/module/ExposureNotificationModule.kt
```diff
@@ -195,6 +199,24 @@ class ExposureNotificationModule(context: ReactApplicationContext) : ReactContex
         }
     }
 
+    @ReactMethod
+    fun provideDiagnosisKeys(diagnosisKeysURLs: ReadableArray, promise: Promise) {
+        promise.launch(this) {
+            try{
+                val files = diagnosisKeysURLs.parse(String::class.java).map { File(it) }
+                val diagnosisKeyFileProvider = DiagnosisKeyFileProvider(files)
+                exposureNotificationClient.provideDiagnosisKeys(diagnosisKeyFileProvider)
```
smcmurtry, line 208: tested locally and `provideDiagnosisKeys` never seemed to return to react native. I added:
```
promise.resolve(null)
```
here and that seemed to fix it.

ios/CovidShield/ExposureNotification.m
```diff
@@ -115,6 +115,25 @@ - (void)invalidate
   }];
 }
 
+RCT_REMAP_METHOD(getExposureWindows, getExposureWindowsFromSummary: (ENExposureDetectionSummary *)summary resolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
```
smcmurtry, line 118: Can we call this `getExposureWindowsFromSummary` rather than `getExposureWindows`? I was running into problems with the type definition calling from react native, since the android fn by the same name does not take an argument.

smcmurtry (approved):
Looks good - we can fix up any bugs in a future PR

jeberhardt:
[Add code to the native layer (Android and iOS) for the V2 API.](https://trello.com/c/gSVf7uWp/682-add-code-to-the-native-layer-android-and-ios-for-the-v2-api)

PR #1235, opened 2020-11-09
V2 Exposure checks on Android

smcmurtry:
# Summary | Résumé

~This is very much a work in progress for now.~ Here is a summary of the changes:
- Added data models to Android native layer for `ExposureWindow`s and `ScanInstance`s
- Added helper functions `toExposureWindow()` and `toScanInstance()`  in native layer to convert objects to the desired type
- Added `getExposureWindowsAndroid` to the adapter to return an array of `ExposureWindow`s from the native layer
- Added a new env variable `USE_V2` to make it easier to turn on the new V2 logic locally
- Updated `ReportType`, `Infectiousness` and `CalibrationConfidence` with the correct integer values from the [documentation](https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/ReportType)
- Added some iOS functions to connect to the native layer that are so far untested


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -429,10 +438,19 @@ export class ExposureNotificationService {
...
       captureMessage('lastCheckedPeriod', {lastCheckedPeriod});
-      await this.exposureNotification.provideDiagnosisKeys(keysFileUrls);
-      const exposureWindows = await this.exposureNotification.getExposureWindows();
+
+      let exposureWindows;
+      if (Platform.OS === 'android') {
+        exposureWindows = await this.exposureNotification.getExposureWindowsAndroid(keysFileUrls);
+      } else {
+        const summaries = await this.exposureNotification.detectExposure(exposureConfiguration, keysFileUrls);
+        exposureWindows = await this.exposureNotification.getExposureWindowsIos(summaries[0]);
```
jeberhardt, line 450: Is it possible that `summaries` is null, or doesn't have a length >=1?

src/env/index.ts
```diff
@@ -27,6 +27,8 @@ export const REGION_JSON_URL = Config.REGION_JSON_URL;
 
 export const EN_CONFIG_URL = Config.EN_CONFIG_URL;
 
+export const EN_API_VERSION = Config.EN_API_VERSION;
```
jeberhardt, line 30: `.env.example` should include an example of EN_API_VERSION with acceptable values.

src/bridge/ExposureNotification/types.ts
```diff
@@ -75,7 +76,7 @@ export interface ExposureConfiguration {
 }
 
 export interface DailySummariesConfig {
-  reportTypeWeights: Record<Report, number>;
+  reportTypeWeights: Record<ReportType, number>;
```
jeberhardt, line 79: I think the `number` type can be removed for this and the next line.

smcmurtry: The type definition for `DailySummariesConfig` is entirely unused and I don't think we'll need it. So I'll remove the whole thing.

src/env/index.ts
smcmurtry: Good call!

src/services/ExposureNotificationService/ExposureNotificationService.ts
smcmurtry: Yes it could have length zero. I'll add a check and will plan to revisit this in a future PR, since I think we should probably be feeding in all of our summaries to `getExposureWindowsIos`.

src/bridge/ExposureNotification/ExposureNotificationAdapter.android.ts
```diff
@@ -33,5 +40,18 @@ export default function ExposureNotificationAdapter(exposureNotificationAPI: Exp
       }
       return [];
     },
+    getExposureWindowsAndroid: async (diagnosisKeysURLs: string[]) => {
+      captureMessage('getExposureWindows');
+      await exposureNotificationAPI.provideDiagnosisKeys(diagnosisKeysURLs);
+      const _exposureWindows = await exposureNotificationAPI.getExposureWindows();
+      const exposureWindows = _exposureWindows.map(window => {
```
jeberhardt, line 47: Having two variables, `_exposureWindows` and `exposureWindws` seems unnecessarily confusing. How about just  returning the results of the `.map()` call instead of creating the 2nd variable?

smcmurtry: Yep good idea.

jeberhardt (approved):

PR #1240, opened 2020-11-10
Clear exposed state

dsamojlenko:
Adds a new button to allow users to clear notifications 

https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=13008%3A15


src/screens/home/views/ExposureDateView.tsx
```diff
@@ -0,0 +1,35 @@
+import React, {useMemo} from 'react';
+import {Text} from 'components';
+import {useI18n} from 'locale';
+import {useExposureNotificationService} from 'services/ExposureNotificationService';
+
+export const ExposureDateView = () => {
+  const i18n = useI18n();
+  const dateLocale = i18n.locale === 'fr' ? 'fr-CA' : 'en-CA';
+
+  const dateFormat =
```
smcmurtry, line 10: Order of the object doesn't matter here - it's determined by the `toLocaleString` function and depends on the locale. 
Screen Shot 2020-11-10 at 12 12 31 PM

timarney:
How to test:

**Quick test**
- Get phone into the exposed state
- Dismiss the alert
- Force a detect using the test menu

**Long test**
- Get phone into the exposed state
- Dismiss the alert
- Wait for background task to run (a few time)

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -346,6 +347,10 @@ export class ExposureNotificationService {
     }
   }
 
+  async clearExposedStatus() {
+    return this.finalize({type: ExposureStatusType.Monitoring});
```
jeberhardt, line 351: What is being returned here? `finalize` doesn't return a value.

jeberhardt, line 351: By only calling `finalize`, is it possible that an artifact like `notificationSent` could remain in the `ExpsosureStatus` value? `notificationSent` would be set to a value when the status is `ExposureStatusType.Exposed`, but I can't find anything that removes that key from the status in `finalize()`.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -483,6 +488,17 @@ export class ExposureNotificationService {
     return true;
   };
 
+  public getExposureDetectedAt(): number | undefined {
+    const exposureStatus = this.exposureStatus.get();
+    let timeStamp;
+
+    if (exposureStatus.type === ExposureStatusType.Exposed && exposureStatus.exposureDetectedAt) {
```
jeberhardt, line 495: It's not necessary to check for `exposureStatus.type === ExposureStatusType.Exposed`.

src/services/ExposureNotificationService/ExposureNotificationService.ts
timarney: It return the promise this happens in `performExposureStatusUpdate` as well

timarney: Good question 

Looking at where the code resets 

`updateExposure` 

After the `EXPOSURE_NOTIFICATION_CYCLE` the `notificationSent` etc.. doesn't get reset.

I'll dig on this a bit.

src/services/ExposureNotificationService/ExposureNotificationService.ts
timarney: TypeScript thing

Screen Shot 2020-11-17 at 9 44 11 AM

Note that block of code has been updated but same comment.

timarney (approved):

PR #1244, opened 2020-11-12
Turn COVID Alert on / off

timarney:
https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=8652%3A4


timarney:
**Google API**

`start()`
> Tells Google Play services to start the broadcast and scan process. The first time that this method is called after installation of the app or each time after stop() is called, it prompts Google Play services to display a dialog box, where the user is asked to give permission to broadcast and scan.


`stop()`
> Disables the broadcast and scan process. You can call this directly. It's also called when a user uninstalls the app. When it’s called as part of the uninstallation process, the database and keys are deleted from the device.

PR #1263, opened 2020-11-25
Feature - log levels and logging improvements

dsamojlenko:
# Summary | Résumé

Introduces a new logging package to help us manage our logging output. This initial PR brings the package in and sets up the transports, but leaves the old `captureMessage` and `captureException` functions in place - they just pass off to the new logger as `info` and `error` levels respectively.

I recommend that we replace all the `captureMessage` and `captureException` calls in a separate PR.

Going forward, instead of importing those methods for logging, we can just import the new logger and then use levels like:

```
import {log} from './shared/logging/config';

log.debug('This is low-level debug stuff that we don't always want to see');
log.info('This is just informational');
log.warn('More important messages');
log.error('Something broke');
```

## Levels
Currently, the following levels are defined:

```
status: 0,
debug: 1,
info: 2,
error: 3,
```

When running the app locally, you can change the target log level to narrow down what's being dumped to console. You can do this with an environment variable called `LOG_LEVEL`. 

You specify the *minimum* level logs you want to see, so `LOG_LEVEL=debug` will output debug, info, error, mad. If you set `LOG_LEVEL=error` you'll only see error and mad. If you set `LOG_LEVEL=silly` you'll see everything.

This will of course improve as we implement appropriate levels to our log calls throughout the code.

## Transports
There are two transports already configured:

### Console
Will output to the console using either `console.info`, `console.warn`, or `console.error` for matching levels, and `console.log` for the rest. There is still some improvements that can be made here like color-coding messages by level, again to be done in a separate PR.

#### Colours
The default Console transport will render logs in the following colours:
- debug: cyan
- info: blue
- error: red
- mad: orange
- other: black

Note that the console in XCode does not render ANSI colours, so if you use that console primarily, you may want to switch to the non-colour console transport.

### Loggly
This transport is only enabled when the LOGGLY_URL is configured in the environment, and the app is not in Production mode. Going forward, we won't need to keep a long-running branch or cherry-pick a commit to enable logging anymore. We can just configure it in our dev, demo, staging environments and be done.

The logs that are sent to Loggly include the level, so we'll be able to use that to filter results in Loggly as well.


src/shared/logging/transports/colour-console.ts
```diff
@@ -0,0 +1,48 @@
+import {transportFunctionType} from 'react-native-logs';
+
+const colourConsoleTransport: transportFunctionType = async (msg, level, _options) => {
+  const colours = {
```
timarney, line 4: Thinking we should add this conditional around the transports as a safeguard

```
if (__DEV__ && !isTest()) {
```

timarney (approved):

PR #1268, opened 2020-11-27
Add check for user stopped - for Unauthorized status panel

timarney:
Fixes iOS 13.6 bug where Status unauthorized panel was showing when Turn off was pressed + the user turned off from iOS settings.

### DIFF
--- a/src/screens/home/views/OverlayView.tsx
+++ b/src/screens/home/views/OverlayView.tsx
@@ -277,7 +277,7 @@ export const OverlayView = ({status, notificationWarning, turnNotificationsOn, b
                 <SystemStatusOff i18n={i18n} />
               </Box>
             )}
-            {status === SystemStatus.Unauthorized && (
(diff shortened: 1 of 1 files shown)


src/screens/home/views/OverlayView.tsx
```diff
@@ -277,7 +277,7 @@ export const OverlayView = ({status, notificationWarning, turnNotificationsOn, b
                 <SystemStatusOff i18n={i18n} />
               </Box>
             )}
-            {status === SystemStatus.Unauthorized && (
+            {!userStopped && status === SystemStatus.Unauthorized && (
```
jeberhardt, line 280: What happens when the user has stopped EN?

timarney: There's an on / off button that take priority + there's an on / off panel in the drawer 

**on / off button** 


**on panel**

jeberhardt (approved):

PR #1283, opened 2020-12-07
Add types to log

timarney:
Add TypeScript definitions to log.error, log.debug etc..

Screen Shot 2020-12-07 at 11 31 45 AM

This helps to ensure we're passing the correct parameters without the need to look them up :)

Note: I'm inline the payload definitions which ends up being repetitive but seems without doing that this seems to be the output (vs the "full preview") above.

Screen Shot 2020-12-07 at 11 51 35 AM


dsamojlenko (approved):
🌮

Issue #1300, opened 2020-12-14
Add support for iOS 12.5 ( iPhone 5s, 6, and 6 Plus)


timarney:
Apple has released iOS 12.5 which includes the Exposure Notifications framework.

https://9to5mac.com/2020/12/14/apple-releases-ios-12-5-for-older-iphones-with-support-for-covid-19-exposure-notifications/

Once updated the app should be able to run on iPhone 5s, 6, and 6 Plus.

**High level updates needed:**
[ ] Set the Deployment Target and Link the Framework
[ ] Perform API and iOS Version Checks
[ ] Set Background Task Interval and Request Runtime to support older iPhones

**Details:**
https://developer.apple.com/documentation/exposurenotification/supporting_exposure_notifications_in_ios_12_5

Ref:
https://github.com/cds-snc/covid-alert-app/issues/1099
https://github.com/cds-snc/covid-alert-app/issues/985

Notes:

> BGTaskScheduler isn’t available in iOS 12.5, Bluetooth explicitly grants apps with the Exposure Notifications entitlement 3.5 minutes at least once a day.

timarney:
https://github.com/cds-snc/covid-alert-app/issues/1099

https://github.com/cds-snc/covid-alert-app/issues/985

dsamojlenko:
Had our first successful exposure test on an iPhone 5S today 🎉 

![image](https://user-images.githubusercontent.com/1187115/102656133-664e0180-4141-11eb-9b4c-863040a804b4.png)

timarney:
Multiple iPhones test 🎉  

<img width="800" src="https://user-images.githubusercontent.com/62242/102692085-0b6ae780-41df-11eb-983b-67802fc2ce1e.jpg" />

timarney:
Further testing has been completed on v6 devices.

A few items remain:

- Finish testing with a second 5s device 
- Error handling for 13.0 -  13.4 devices

> Although iOS 13.0 to 13.4 doesn’t support Exposure Notifications, your app might still download onto devices with those versions of iOS. If that happens, present a dialog to let the user know Exposure Notifications isn’t available for their device.

timarney:
Merged
https://github.com/cds-snc/covid-alert-app/pull/1306

PR #1306, opened 2020-12-16
Add support for iOS 12.5

dsamojlenko:
# Summary | Résumé

Updates app to support iPhone 5s, 6, and 6 Plus

Related: #1300 

Adds support for iOS 12.5. 

- [X] Set the Deployment Target and Link the Framework
- [x] Perform API and iOS Version Checks
- [x] Set Background Task Interval and Request Runtime to support older iPhones
- [x] Handle iOS version above 12.5 and below 13.5 as they do not support Exposure Notifications framework

**Testing**
- [x] End to end exposure test
- [x] Daily logs


timarney:
It's running :) 

Nice work @craigzour @dsamojlenko

ios/CovidShield/ExposureNotification.m
```diff
@@ -130,18 +130,13 @@ - (void)invalidate
...
-        reject([NSString stringWithFormat:@"%ld", (long)error.code], error.localizedDescription ,error);
-      } else {
-        resolve(exposureWindows);
-      }
-    }];
-  } else {
-    // Fallback on earlier versions
-    reject(@"API_NOT_AVAILABLE", @"API Not Available. Requires iOS 13.7+", nil);
-  }
+  [self.enManager getExposureWindowsFromSummary:(ENExposureDetectionSummary *)summary completionHandler:^(NSArray<ENExposureWindow *> * _Nullable exposureWindows, NSError * _Nullable error) {
```
craigzour, line 181: To double check before approving this change

dsamojlenko:
image

Background Exposure Check success!

timarney:
iPhone 6 (successful exposure)

timarney:
Adding this screenshot for reference:

src/screens/landing/LandingScreen.tsx
```diff
@@ -10,13 +10,41 @@ export const LandingScreen = () => {
...
+
+      isENFrameworkSupported()
+        .then(isSupported => {
+          if (isSupported) {
+            navigation.reset({
+              index: -1,
+              routes: [{name: 'OnboardingNavigator'}],
+            });
+          } else {
+            Alert.alert(
```
timarney, line 39: 1) This will need to go in the locale files - if your not sure how to do that I can help.

2) We'll need to ensure outside of the Alert the user would land on the "something went wrong" homescreen if they make it that far.

craigzour: This commit is just a technical prototype. The way we want to transmit the information through the UI should be something defined by the product team. Do you agree @timarney ?

aklef: > 2. We'll need to ensure outside of the Alert the user would land on the "something went wrong" homescreen if they make it that far.

This can easily happen if the app was previously installed on a device

craigzour: Screen Shot 2020-12-29 at 4 06 53 PM
Found a view that was already implemented in the code. It is already translated in both French and English. Will redirect the user to that screen instead of displaying the alert.

timarney (approved):

PR #1320, opened 2020-12-22
Prevent exposure checks when EN Framework is turned off on Android.

### LINKED ISSUE #1315: Deactivated app still showing exposure check notice
# Summary | Résumé

As a user living in BC (or Alberta), I deactivated my app (in order to minimize battery drain for zero gain) and did notice a day later this message zipping by telling me the app was checking for exposure. It may not be a real problem (as I can just ignore message) but it does not inspire confidence. In addition, I have had no contacts worth checking the exposures for. Seems highly inefficient to check against exposures that were reported in other provinces, certainly when travel is discouraged and people are expected to isolate. Therefore check is draining server and app resources unnecessarily as well.  Worse when app is deemed deactivated. This adds up with millions of users. There should be much better ways to bring app to foreground so it does not hibernate, such as issuing a daily notice on proximity contacts log under #1264. Usefulness of app would also be better demonstrated. Late bug report addition: it was also noticed that the "Last checked for potential exposure" datestamp logged under the Google setting account reporting page did not match the latest check message zooming by, it was in fact 5 days older. 

# Steps to reproduce
Disable app using new menu option. Wait for message to appear. Took a day before I noticed.

# Environment | Environnement logiciel
Android latest app release

Should there be exposure checks in the first place if there are no proximity contacts whatsoever? Perhaps the exposure check was performed for the period before I deactivated (for which there might have been odd contacts - can't tell since contact daily log not implemented yet)??? It would make more sense to do these checks only when app is activated, or reactivated. Unless you really mean to check on these health worker folks even when they deactivate their app. 

> Are there any related issues you consider out of scope for this issue that could be addressed in the future?
None

jeberhardt:
# Summary | Résumé

When running background tasks on Android, if the EN framework is deactivated, a visual notification briefly appears indicating that an Exposure Check is happening, even though no check is happening.

This conditional check happens at the beginning of the background task in Kotlin, prior to the execution of the code being passed off to react-native.

Closes #1315

### DIFF
--- a/android/app/src/main/java/app/covidshield/receiver/worker/ExposureCheckSchedulerWorker.kt
+++ b/android/app/src/main/java/app/covidshield/receiver/worker/ExposureCheckSchedulerWorker.kt
@@ -8,7 +8,10 @@ import app.covidshield.extensions.log
 import com.facebook.react.ReactApplication
 import com.facebook.react.bridge.ReactContext
 import com.facebook.react.modules.core.RCTNativeAppEventEmitter
+import com.google.android.gms.nearby.Nearby
+import com.google.android.gms.nearby.exposurenotification.ExposureNotificationStatus
 import kotlinx.coroutines.*
+import kotlinx.coroutines.tasks.await
 
 private const val HEADLESS_JS_TASK_NAME = "EXPOSURE_CHECK_HEADLESS_TASK"
 private const val HEADLESS_JS_TASK_TIMEOUT_MS = 60000L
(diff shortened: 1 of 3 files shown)


jeberhardt:
[Exposure check notifications should not appear when framework is turned off](https://trello.com/c/gPkmjUWV/925-exposure-check-notifications-should-not-appear-when-framework-is-turned-off)

android/app/src/main/java/app/covidshield/receiver/worker/ExposureCheckSchedulerWorker.kt
```diff
@@ -8,17 +8,29 @@ import app.covidshield.extensions.log
...
 
+    private val exposureNotificationClient by lazy {
+        Nearby.getExposureNotificationClient(context)
+    }
+
     override suspend fun doWork(): Result {
         Log.d("background", "ExposureCheckSchedulerWorker - doWork")
         try {
+            val enStatus = exposureNotificationClient.status.await()
+            if (!enStatus.contains(ExposureNotificationStatus.ACTIVATED)){
```
timarney, line 30: Should we use `INACTIVATED` here?

> Exposure notification is not running

That would avoid checking other possible states i.e. the UNKNOWN status etc...

timarney: This use isEnabled here

Found this:

https://developers.google.cn/android/exposure-notifications/implementation-guide?hl=zh-cn#doWork

```

    override suspend fun doWork(): Result {
       ...

        return try {
            // Skip work if API is not enabled or the Worker has been stopped.
            if (isEnabled && !isStopped()) {
              ...
```

timarney: Also noting for isEnabled

> Note: This method does not tell you whether the system is actually scanning or not. In other words, this returns true even if BT is off.

jeberhardt: @timarney - strangely enough, the reference app from Google doesn't seem to implement `isDisabled` at all.

timarney: Looking to use `isEnabled`

jeberhardt: Ha ha! I meant to type `isEnabled` in my comment above.  ;)

timarney (approved):

PR #1325, opened 2020-12-30
Feature/share exposures - PT 1

omartehsin1:
Part 1 - `Adds Main Screen`

Follow up PRs to continue the rest of the full as outlined below.

## Full Description
10% of people who have entered a OTK have not uploaded their codes.

## Motivation
When someone has entered their OTK but has not uploaded their codes
They need to know they are not finished their task of letting people know they've been exposed
And they need to know what to do next
So that their close contacts will receive an exposure notification

## Acceptance criteria

**Main screen**

Given people have entered their OTK but have not uploaded their codes
When they leave the flow or otherwise land on the main page of the app
Then they see messaging that they are not quite done ([design](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A26))

**Menu**

Given people have entered their OTK but have not uploaded their codes
When they open the menu
Then they see the prompt to share exposures ([design](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A52))

**Intermediate instructions screen**

Given people have entered their OTK but have not uploaded their codes
When they click "Finish sharing exposures"
Then they see the intermediate screen with steps, and the first step is now "Done" ([design](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A740))
And when they click "Next"
Then they follow the rest of the existing process (entering date of symptom onset, etc)

**Confirmation of completed the flow**

When people have entered their OTK and shared their exposures, completing the flow
Then their screen says "Thanks for helping fight COVID-19" and tells them to share any new exposures going forward ([text change - design)](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A521)

**Sharing exposures on subsequent days**
Given someone has entered a OTK and shared exposures on a previous day
When they open their app
Then they see the "Share new exposures screen" ([text change - design](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A1112)).


ios/CovidShield.xcodeproj/project.pbxproj
```diff
@@ -561,7 +561,7 @@
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = ca.gc.hcsc.canada.stopcovid;
 				PRODUCT_NAME = CovidShield;
-				PROVISIONING_PROFILE_SPECIFIER = "Stop COVID Development Profile";
```
timarney, line 564: This file should be reverted - not part of this PR

src/screens/home/HomeScreen.tsx
```diff
@@ -109,8 +110,13 @@ const Content = ({isBottomSheetExpanded}: ContentProps) => {
       if (!network.isConnected) {
         return <NetworkDisabledView />;
       }
+
       return exposureStatus.needsSubmission ? (
-        <DiagnosedShareView isBottomSheetExpanded={isBottomSheetExpanded} />
+        exposureStatus.hasShared ? (
```
timarney, line 115: Thinking either force hasShared to always be false for this PR or always default to the existing DiagnosedShareView for this PR so that we can merge to Master without breaking existing functionality (before the rest of the pieces for this flow are done).  

You can add back in a follow-up PR after.

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -758,7 +760,7 @@ export class ExposureNotificationService {
   private async recordKeySubmission() {
     const currentStatus = this.exposureStatus.get();
     if (currentStatus.type !== ExposureStatusType.Diagnosed) return;
-    this.exposureStatus.append({needsSubmission: false, submissionLastCompletedAt: getCurrentDate().getTime()});
+    this.exposureStatus.append({needsSubmission: false, hasShared: true, submissionLastCompletedAt: getCurrentDate().getTime()});
```
timarney, line 763: Per the other comment if you force this to false for this PR we could merge without breaking existing functionality :) and you could keep the other logic.

src/screens/home/HomeScreen.tsx
omartehsin1: So the `DiagnosedShareUploadView` screen appears when `hasShared` is false (because use has not shared the upload)  so changing it to true will take user to the `DiagnosedShareView `, even if they have not finished uploading (instead of false which will do the opposite). That work?

timarney: Yes that works but we don't want that to happen until the rest of the functionality for this new flow is complete.

If we can adjust and merge this PR and than re-introduce (hasShared) we will avoid the possibility of a user hitting the new screen in the meantime.

timarney (approved):

timarney (approved):

PR #1338, opened 2021-01-06
Get e2e tests running again

dsamojlenko:
# Summary | Résumé

Some effort was invested setting up Detox e2e tests a while ago, but the initial tests are out of date, and the test suite is currently not able to run. This PR will get them running again, follow up PRs will get the old tests fixed or removed, and new tests written.

First issue we ran into was for iOS, the xcodebuild command would fail when building the simulator to be used for testing. Investigation lead us to believe this was being caused by arm64 architecture support introduced in Xcode 12. [See this StackOverflow discussion](https://stackoverflow.com/questions/63607158/xcode-12-building-for-ios-simulator-but-linking-in-object-file-built-for-ios/63955114#63955114)

When trying to get Android tests running, we ran into some Java version issues with the AOSP emulators. The Readme has been updated with additional instructions for dealing with this.


aklef (approved):
Looks good and I managed to get setup with running tests on both an android and ios emulator

PR #1342, opened 2021-01-06
Configuration files for i18n-ally extension for VS Code.

jeberhardt:
# Summary | Résumé

This is a modification for use in development only. It has no impact on the released app.

Adding configuration files for use with **VS Code** for the i18n-ally extension: 

The extension will help:
-- identify strings that haven't been translated in any language
-- show in-line copy of the actual string
-- have a side-by-side editor for all languages of the string in question
-- show lint warnings for any strings in the code that can't be found in the language files
-- plus some more stuff
-- show all translations when hovering over the code

Install the i18n-ally extension: https://marketplace.visualstudio.com/items?itemName=antfu.i18n-ally

Open the project in VS Code, and look for the "globe" icon on the left-hand side. Clicking on that should open the extension. Check the documentation for the extension for further details.

Should probably move the `region.json` file, since it's not translated in the same way. Also move other files that are not language specific JSON files.

Screen Shot 2020-12-23 at 5 16 11 PM

### DIFF
--- a/.vscode/i18n-ally-custom-framework.yml
+++ b/.vscode/i18n-ally-custom-framework.yml
@@ -1,0 +1,19 @@
+# .vscode/i18n-ally-custom-framework.yml
+
+# An array of string which contains Language Ids defined by vscode
+# You can check avaliable language ids here: https://code.visualstudio.com/docs/languages/overview#_language-id
+languageIds:
+  - javascript
+  - typescript
+  - javascriptreact
+  - typescriptreact
+
+# An Array of regex to find the keys usage. **The key should captured in the first match group**.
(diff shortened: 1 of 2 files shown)


timarney:
> Should probably move the region.json file, since it's not translated in the same way. Also move other files that are not language specific JSON files.

I'm looking to split region.json into multiple lang files ... with the current region.json file be a generated file (via a build script).  The generated file would need to be copied to the server repos at least to start with.

I'll likely get to that next week.

timarney (approved):
This is really great.  Will save a bunch of time doing string lookups.

🎉

PR #1346, opened 2021-01-07
Feature/otk notification

omartehsin1:
Part 3 - `OTK notification`

Follow up PRs to continue the rest of the full as outlined below.

## Full Description
10% of people who have entered a OTK have not uploaded their codes.

## Motivation
When someone has entered their OTK but has not uploaded their codes
They need to know they are not finished their task of letting people know they've been exposed
And they need to know what to do next
So that their close contacts will receive an exposure notification

## Acceptance criteria

**Main screen**

Given people have entered their OTK but have not uploaded their codes
When they leave the flow or otherwise land on the main page of the app
Then they see messaging that they are not quite done ([design](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A26))

**Menu**

Given people have entered their OTK but have not uploaded their codes
When they open the menu
Then they see the prompt to share exposures ([design](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A52))

**Intermediate instructions screen**

Given people have entered their OTK but have not uploaded their codes
When they click "Finish sharing exposures"
Then they see the intermediate screen with steps, and the first step is now "Done" ([design](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A740))
And when they click "Next"
Then they follow the rest of the existing process (entering date of symptom onset, etc)

**OTK Notification**
If people have only entered the OTK but have not shared, they will receive a notification reminding them to complete it.
(https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A74)

**Confirmation of completed the flow**

When people have entered their OTK and shared their exposures, completing the flow
Then their screen says "Thanks for helping fight COVID-19" and tells them to share any new exposures going forward ([text change - design)](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A521)

**Sharing exposures on subsequent days**
Given someone has entered a OTK and shared exposures on a previous day
When they open their app
Then they see the "Share new exposures screen" ([text change - design](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=14333%3A1112)).

------------------------------------------------------------------------------------------------------------------------
This PR includes the push notification that users will get if they close the app or turn off the screen before they've completed uploading their OTK. 

[...]

PR #1350, opened 2021-01-12
Check if uploading

timarney:
Patch fix for https://github.com/cds-snc/covid-alert-app/pull/1346

Adds check to see if the user is in the upload flow - this will be used to prevent a "not quite done" notification if the user is still in the upload flow.   Issue found where the app would move into `the background` when the share screen prompt appears vs the user closing the flow and sending he app to the background. 

Android
1. Enter OTK, go through share exposure flow (date or not date route)
2. Tap share exposures
3. Observe the OS prompt to “Share your random IDs with COVID Alert” AND a notification doesn't appear i.e. "you’re not quite done"

### DIFF
--- a/src/screens/datasharing/components/BaseTekUploadView.tsx
+++ b/src/screens/datasharing/components/BaseTekUploadView.tsx
@@ -31,7 +31,7 @@ export const BaseTekUploadView = ({
   const navigation = useNavigation();
   const i18n = useI18n();
   const [loading, setLoading] = useState(false);
-  const {fetchAndSubmitKeys, isUploading} = useReportDiagnosis();
+  const {fetchAndSubmitKeys, setIsUploading} = useReportDiagnosis();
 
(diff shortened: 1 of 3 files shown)


src/screens/datasharing/components/BaseTekUploadView.tsx
```diff
@@ -71,12 +71,16 @@ export const BaseTekUploadView = ({
   };
   const handleUpload = useCallback(async () => {
     setLoading(true);
+    isUploading(true);
+
     try {
       await fetchAndSubmitKeys(contagiousDateInfo);
       setLoading(false);
+      isUploading(false);
```
smcmurtry, line 79: I would find this a bit easier to read if the function was named `setIsUploading(false)`

smcmurtry (approved):
Tested via a local build and I am no longer seeing the premature "you're not done yet" notification. One minor comment.

PR #1316, opened 2020-12-21
Feature/multiple exposure 

timarney:
https://trello.com/c/HqyhF3aK/806-people-can-receive-notifications-when-in-the-exposed-state

https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=13538%3A14

PR #1355, opened 2021-01-12
Add an enum for the `forceScreen` variable

smcmurtry:
# Summary | Résumé

This is in response to a bug in #1316 where we checked that no screen was forced with:
```
if (forceScreen) {
// assume that user is forcing a screen
}
```

This was a bug because `forceScreen === 'None'` when the user makes a selection and then resets the option. Using an enum will reduce the chances we will make this mistake again.

### DIFF
--- a/src/screens/home/HomeScreen.tsx
+++ b/src/screens/home/HomeScreen.tsx
@@ -10,15 +10,15 @@ import {
   useExposureStatus,
   useUpdateExposureStatus,
   useStartExposureNotificationService,
+  useExposureNotificationSystemStatusAutomaticUpdater,
   useSystemStatus,
 } from 'services/ExposureNotificationService';
 import {useStorage} from 'services/StorageService';
+import {RegionCase} from 'shared/Region';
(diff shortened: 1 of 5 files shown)


src/screens/home/HomeScreen.tsx
```diff
@@ -10,16 +10,16 @@ import {
...
 } from 'services/ExposureNotificationService';
 import {useStorage} from 'services/StorageService';
+import {RegionCase} from 'shared/Region';
 import {getRegionCase} from 'shared/RegionLogic';
 import {usePrevious} from 'shared/usePrevious';
+import {ForceScreen} from 'shared/ForceScreen';
 import {useRegionalI18n} from 'locale';
 
-import {useExposureNotificationSystemStatusAutomaticUpdater} from '../../services/ExposureNotificationService';
-import {RegionCase} from '../../shared/Region';
```
smcmurtry, line 21: This is an unrelated change - just thought I'd clean up the imports a bit

timarney (approved):

timarney:
🎉

PR #1361, opened 2021-01-14
Generate region.json

timarney:
# Summary | Résumé

Adds a script + yarn command to generate the `region.json` file

Contents of the existing `region.json` have been moved to a new regional content folder and split up by language.

The script will combine the contents of the locale/translations/regional (json files) and write the content to the existing `region.json`

---

## Test 1:

It generates a file

1) Run `yarn generate-translations:regional` note the output of `region.json`

<hr>

## Test 2:

Testing the content `region.json` vs new generated `region.json`

1) Take contents from current (master) `region.json` and minify the content i.e. https://codebeautify.org/jsonminifier

2) Generate the new content `yarn generate-translations:regional` + minify https://codebeautify.org/jsonminifier

3) Diff the 2 files i.e. https://www.diffchecker.com/


smcmurtry (approved):
Looks good! I ran the tests and the minified region.json files match up. One thing I noticed - the region.json file the script outputs is already minified so it doesn't need to be minified again.

PR #1362, opened 2021-01-14
Testing/fix existing e2e tests

dsamojlenko:
# Summary | Résumé

Get existing e2e tests to passing


aklef (approved):
Tested on both android and iOS and test pass, looks good

PR #1372, opened 2021-01-18
Improve Deployment Scripts

dsamojlenko:
# Summary | Résumé

Simplifies our Fastlane deployment scripts by making use of Xcode Build Configurations and Schemes. 

Previously, Fastlane had been attempting to set Bundle Identifier and Provisioning Profile (among other settings) at build time, but this was always a bit flaky and required some manual attention before starting a build.

This PR fixes that issue by adding Build Configurations and Schemes for each applicable environment, which take care of setting those properties.

Another related improvement is reverting to using placeholders in .plist and project files for Version Name/Code. This saves us from force-applying the change to these files and having dirty files after a build.

Other minor changes:
- Simplified APP_ID property
- Use env vars for Profile, Scheme, Template
- Remove prompt at beginning to check profile/bundle id
- Updated .env file example


fastlane/iOS.Fastfile
```diff
@@ -1,29 +1,14 @@
...
 
     versions = Array(testflight_latest)
 
     UI.user_error!("Version code #{versions} has already been used!") if Array(versions).include? ENV['APP_VERSION_CODE'].to_i
   end
 
-  # Saving this for later, but may not need anymore
-  lane :set_version do
-    # Set the version name from the environment
-    increment_version_number(
```
timarney, line 15: After reviewing with @dsamojlenko looks like we need to put this back - but we'll test further.

dsamojlenko: Couldn't find another fix, so brought this code back

timarney (approved):

PR #1374, opened 2021-01-19
fixed issue with overlay menu

omartehsin1:
# Summary | Résumé

There was an issue with the overlay screen that led the menu being 'frozen', users would not be interact with any of the buttons on the menu. This was caused by conflicting bottom sheet booleans. In the `HomeScreen` , the `bottomSheetRef` was calling the `collapse()`, setting `bottomSheet.isExpanded` to false, however, the `isBottomSheetExpanded`, had a value of true. This is why the Overlay Menu was appearing, but could not be interacted with. Removed the function because it only effected the bottomSheet in diagnosed mode which led to the issue.

### DIFF
--- a/src/screens/home/HomeScreen.tsx
+++ b/src/screens/home/HomeScreen.tsx
@@ -16,7 +16,6 @@ import {
 import {useStorage} from 'services/StorageService';
 import {RegionCase} from 'shared/Region';
 import {getRegionCase} from 'shared/RegionLogic';
-import {usePrevious} from 'shared/usePrevious';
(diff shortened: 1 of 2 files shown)


timarney:
@smcmurtry please give this a review :)

smcmurtry (commented):
So this just removes the behaviour where the bottom closes after completing the diagnosis flow. I think this is an important feature to keep, since it means that the user sees the message on the home screen right away, either: "thank you" or "not quite done".

What are the steps to reproduce the bug this is trying to solve?

timarney:
Closing this - believe this ended up being a non issue.

PR #1373, opened 2021-01-19
Add metrics hook

timarney:
# Summary | Résumé

Initial work to add metrics hooks

This is the UI piece which will connect to https://github.com/cds-snc/covid-alert-app/pull/1371

Current logging is for dev mode only 
```
raw message
{"uuid":"S7GSNZ8T","platform":"android","msg":{"category":"metrics","message":"en-toggle","payload":{"appVersion":209,"appOs":"android","payload":{"timestamp":1611250306668,"identifier":"en-toggle","region":"ON","data":{"state":false}}}},"level":{"severity":5,"text":"debug"},"versionCode":209,"versionName":"1.1.7","currentStatus":"monitoring","lastCheckedMinutesAgo":"1"}
```

## Testing

Install the app 
Onboard 
Toggle EN on/off vs the drawer menu or home screen or info panel 

You should now be able to events for install, onboarded and multiple en toggle events depending on what you press

PR #1378, opened 2021-01-21
Clear exposed metric

timarney:
Add clear exposed metric

Follow up for https://github.com/cds-snc/covid-alert-app/pull/1373

Adds utility function to calculate hours between exposed to clearing

```
getHoursBetween(...)
```


src/shared/metrics.tsx
```diff
@@ -47,6 +48,10 @@ interface EnToggleMetric extends Metric {
   state: boolean;
 }
 
+interface ClearExposedMetric extends Metric {
+  timeamount: number;
```
smcmurtry, line 52: I think this should have the units in it, maybe `hoursSinceExposureDetectedAt` or something along those lines.

smcmurtry (approved):
Looks good! One minor comment.

PR #1383, opened 2021-01-22
Feature/expand e2e tests

aklef:
# Summary | Résumé

@dsamojlenko recently revived the Detox e2e tests in PR #1338 In this follow up PR, a new test is created to validate features of the app's internal Demo menu.

Note: testing the sample notification is a tricky feat, as the notification popup is a system-generated element and not part of our app. Thus, in Detox tests we can't easily apply an ID handle to it, which would be needed to easily reference the popup among the many elements that make up the UI stack.. 

On ios, we could probably do something weird like matching "by.type('_UIAlertControllerActionView')" but further research is required for this, and for android..

Also, I had no issues using an iOS Emulator for screenshots, despite some of our old docs saying the contrary.


aklef:
[Create scripted end to end tests with Detox](https://trello.com/c/MGVmVx5G/947-create-scripted-end-to-end-tests-with-detox)

aklef:
I want to note that these tests seems broken on android for whatever reason... putting this on the backburner as we attempt to upgrade the Detox version we're running

aklef:
Work here is done! Ended up not moving to v18 as lift was too large. going to move on for now and get this merged soon

dsamojlenko (approved):
Noting that the swiping seems to be fairly inconsistent across environments, so maybe something to be improved in future.

PR #1384, opened 2021-01-25
QR code prototype

timarney:
# Summary | Résumé

For details see:
https://github.com/cds-snc/covid-alert-documentation/blob/main/product/qr-codes.md


src/screens/qr/CheckInSuccessfulScreen.tsx
```diff
@@ -0,0 +1,68 @@
...
+    </BaseDataSharingView>
+  );
+};
+
+function urlToString(url: string): string {
+  const title = url.replace(/_/g, ' ');
+  return title;
+}
+
+const returnTodaysDate = (): string => {
```
timarney, line 57: @smcmurtry is there something we already have for this in Date FNS?

src/screens/qr/components/Toolbar.tsx
```diff
@@ -0,0 +1,61 @@
...
+      {showCloseButton ? (
+        <Box style={styles.right}>
+          <Button testID="toolbarCloseButton" text={navText} variant="text" onPress={onIconClicked} />
+        </Box>
+      ) : null}
+    </Box>
+  );
+};
+
+export const ToolbarWhite = ({isWhite}: {isWhite?: boolean}) => {
```
smcmurtry, line 30: The name `ToolbarWhite` makes me think this is the white toolbar, not that it will switch. Maybe just call it `QRToolbar`? Also the prop could be called `useWhiteText` so it's obvious what is changing colour.

src/screens/qr/components/Toolbar.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const navigation = useNavigation();
+  const i18n = useI18n();
+  return isWhite ? (
+    <BackButton
+      textStyles={styles.backText}
+      iconName="icon-chevron-back-white"
+      text={i18n.translate(`QRCode.Reader.Back`)}
+      onPress={() => {
+        navigation.navigate('Home');
```
smcmurtry, line 39: This is the back button, so the function should be `navigation.goBack()`. Also it should be wrapped in a `useCallback` so that the function doesn't have to be re-assigned to the variable whenever the component re-renders.

src/screens/qr/components/Toolbar.tsx
```diff
@@ -0,0 +1,61 @@
...
+      }}
+    />
+  ) : (
+    <Button
+      backButton
+      text={i18n.translate(`QRCode.Reader.Back`)}
+      variant="text"
+      onPress={() => {
+        navigation.navigate('Home');
+      }}
```
smcmurtry, line 49: Same comment as above.

src/screens/qr/components/Toolbar.tsx
```diff
@@ -0,0 +1,61 @@
+import React, {useCallback} from 'react';
+import {StyleSheet} from 'react-native';
+import {Box, Button} from 'components';
+import {useNavigation} from '@react-navigation/native';
+import {useI18n} from 'locale';
+import {BackButton} from '../views/BackButton';
+
+export interface ToolbarProps {
+  onIconClicked(): void;
```
smcmurtry, line 10: We also use this prop name on the datasharing toolbar but I think it is vague. Better would be `onClose`

src/screens/qr/components/BaseQRCodeScreen.tsx
```diff
@@ -0,0 +1,56 @@
...
+  showBackButton?: boolean;
+  showCloseButton?: boolean;
+}
+
+export const BaseQRCodeScreen = ({children, showBackButton, showCloseButton}: BaseQRCodeScreenProps) => {
+  const navigation = useNavigation();
+  const i18n = useI18n();
+  const close = useCallback(() => navigation.navigate('Home'), [navigation]);
+  return (
+    <KeyboardAvoidingView style={styles.flex} behavior={Platform.OS === 'ios' ? 'padding' : 'height'}>
```
smcmurtry, line 21: I don't think we need the `KeyboardAvoidingView` here. The purpose of this is to move things in the View when the device keyboard pops up. We're using this on the DataSharing flow because the OTK entry page, but there is nothing  in the QR Code flow that pops the keyboard.

src/screens/qr/CheckInSuccessfulScreen.tsx
smcmurtry: No - we have `formatExposedDate` but that format doesn't include the time. We could add this format to shared/date-fns.ts.

src/screens/qr/views/NoCamera.tsx
```diff
@@ -15,36 +16,34 @@ export const NoCamera = ({updatePermissions}: {updatePermissions: () => void}) =
...
-            </Text>
+    <BaseQRCodeScreen showBackButton showCloseButton={false}>
+      <Box marginBottom="m">
+        <Toolbar
+          title=""
+          navText={i18n.translate('RegionPicker.Close')}
+          navLabel={i18n.translate('RegionPicker.Close')}
+          navIcon="icon-back-arrow"
+          onIconClicked={close}
+        />
```
smcmurtry, line 27: I think this can be deleted since it is already in the `BaseQRCodeScreen`?

src/screens/qr/views/QRCodeScanner.tsx
```diff
@@ -45,7 +44,7 @@ export const QRCodeScanner = () => {
     >
       <View style={styles.layerTop} />
       <Box style={styles.back} paddingHorizontal="m">
-        <ToolbarWhite isWhite={true} />
+        <ToolbarWhite isWhite />
```
smcmurtry, line 40: If we're not going to use `BaseQRCodeScreen` on the only screen that needs the white text toolbar, let's remove the `isWhite` prop from `BaseQRCodeScreen` and hard code it to use black text on the toolbar.

src/screens/qr/views/NoCamera.tsx
```diff
@@ -15,36 +16,34 @@ export const NoCamera = ({updatePermissions}: {updatePermissions: () => void}) =
...
+      </Box>
+      <ScrollView style={styles.flex}>
```
smcmurtry, line 29: Also I think this can be deleted since there is already a `ScrollView` in `BaseQRCodeScreen`

smcmurtry (approved):

PR #1388, opened 2021-01-27
Patch to fix disappearing exposure history

smcmurtry:
# Summary | Résumé

Removing metrics from ExposureNotificationService temporarily since this is causing a new instance of the class to be created whenever bluetooth toggles on or off.

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationServiceProvider.tsx
+++ b/src/services/ExposureNotificationService/ExposureNotificationServiceProvider.tsx
@@ -8,7 +8,6 @@ import SystemSetting from 'react-native-system-setting';
 import {ContagiousDateInfo} from 'shared/DataSharing';
 import {useStorage} from 'services/StorageService';
 import {log} from 'shared/logging/config';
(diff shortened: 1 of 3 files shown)


src/services/ExposureNotificationService/ExposureNotificationServiceProvider.tsx
```diff
@@ -50,9 +48,8 @@ export const ExposureNotificationServiceProvider = ({
         storage || AsyncStorage,
         secureStorage || RNSecureKeyStore,
         exposureNotification || ExposureNotification,
-        metrics.service,
       ),
-    [backendInterface, exposureNotification, i18n, metrics.service, secureStorage, storage],
```
smcmurtry, line 55: This was the line causing the bug - `metrics.service` must be changing whenever bluetooth is toggled on/off, causing a new instance of `ExposureNotificationService` to be created.

smcmurtry:
Update - this fixes 2 other bugs:
- Diagnosed/upload always leading to the reminder screen
- EN popup re-appearing continuously when bluetooth is off

timarney (approved):

PR #1397, opened 2021-02-01
Remove what's new box

timarney:
# Summary | Résumé

Remove the What's New Box from that was used for an older Android update.


smcmurtry (changes requested):
It also needs to be removed from `NoExposureNoRegionView`

smcmurtry (approved):
Looks good!

PR #1407, opened 2021-02-09
Home screen experiment

smcmurtry:
I changed the home screen so it is made up of 2 panels - the exposure status on top, and the system status below. This is empty if the system status is fine.


timarney:
@smcmurtry any reason to keep this open?

timarney:
Closing @smcmurtry we can re-open as need be.

PR #1412, opened 2021-02-11
Added a new env variable QR_ENABLED

smcmurtry:
# Summary | Résumé

This is because `QR_HOST` will need to always be set, so it would be best not to use it to determine if the QR code scan button is displayed in the app.


omartehsin1 (approved):
Nice, no longer seeing the scan button

timarney (approved):

PR #1414, opened 2021-02-12
Fix ios builds (and update fastlane)

dsamojlenko:
# Summary | Résumé

Fixes a reference to an old build config that was changed in #1372


timarney:
🎈  🥇

timarney (approved):

PR #1417, opened 2021-02-15
able to check in foreground

omartehsin1:
# Summary | Résumé

Initial Work for outbreak exposure


src/screens/home/HomeScreen.tsx
```diff
@@ -218,10 +219,11 @@ export const HomeScreen = () => {
   const startAndUpdate = useCallback(async () => {
     if (userStopped) return;
     const success = await startExposureNotificationService();
+    checkForExposures();
```
smcmurtry, line 222: This means the outbreak checks won't happen if the user has turned the app off. Good call.

smcmurtry: Can you put this inside an `if (QR_ENABLED) {...}` check? Might cause an error if the feature isn't turned on.

smcmurtry (approved):
Looks good!

PR #1419, opened 2021-02-15
Verify QR code

timarney:
Adds signed verification to QR code scanning.


smcmurtry (approved):
Looks good! With the correct `QR_CODE_PUBLIC_KEY` set I can scan the QR no problem. With a bad public key set I get the "QR code not recognized" screen.

PR #1423, opened 2021-02-16
Skip sign qr

timarney:
# Summary | Résumé

Skip QR code verification step if .env var doesn't exist

---


smcmurtry (approved):
Tested locally both with `QR_CODE_PUBLIC_KEY` set and not set. Works great!

PR #1426, opened 2021-02-18
Internal storage API revamp

craigzour:


craigzour:
Now going through initial testing phase :)

feralbob:
Hi @craigzour 

Does any of the new implementation support anything with Keychain Access groups on iOS? 
I'm looking to support sharing between widgets and the app and hoping to find a way to reuse existing code rather than have to write up a custom widget service with iOS/Android implementations. Disclaimer - I have yet to look at codepaths for how widgets on Android can access data from the host app

craigzour:
Hi @feralbob 

The new implementation just provides a unique and centralized wrapper on top of two storage libraries (https://github.com/pradeep1991singh/react-native-secure-key-store ; https://github.com/react-native-async-storage/async-storage) that were already used by the application. The RNSecureKeyStore provides a way to define an accessibility level but we are only using the `ALWAYS_THIS_DEVICE_ONLY` option so I am not sure if that helps.

feralbob:
> Hi @feralbob
> 
> The new implementation just provides a unique and centralized wrapper on top of two storage libraries (https://github.com/pradeep1991singh/react-native-secure-key-store ; https://github.com/react-native-async-storage/async-storage) that were already used by the application. The RNSecureKeyStore provides a way to define an accessibility level but we are only using the `ALWAYS_THIS_DEVICE_ONLY` option so I am not sure if that helps.

Ok thanks for the info.

Relevant to this might be #1444 - I have a draft PR #1445 that updates from @react-native-community/async-storage to @react-native-async-storage/async-storage - I don't see that this PR actually makes that dependency change.

craigzour:
PR has been rebased on top of the latest master. Going through initial testing

src/services/StorageService/StorageService.ts
```diff
@@ -1,124 +1,82 @@
...
+      }
+    });
+  }
 
-    const qrEnabled = (await AsyncStorage.getItem(Key.QrEnabled)) === '1';
-    this.qrEnabled.set(qrEnabled);
-  };
+  deteleAll(): Promise<void> {
+    return this.serialPromiseQueue.add(() => {
+      return this.unsecureKeyValueStore.deleteAll();
```
jeberhardt, line 79: Should this also include something like `this.secureKeyValueStore.deleteAll()`

craigzour: Unfortunately there is no existing API in RNSecureKeyStore to clear the storage. This feature is only accessible to QA through the debug menu and it looks like it is enough for them to reset the state of the app at this point. Definitely something we need to keep in mind if we start using the secure storage to keep `state` type of information.

craigzour:
Integrated @feralbob 's PR https://github.com/cds-snc/covid-alert-app/pull/1446

PR #1427, opened 2021-02-18
Homescreen z-index test

amazingphilippe:
# Summary | Résumé

Just a test using z-index to pull the shape on top of the first card. The negative margin is set on the rounded box, only if its the first one. Nothing else is added to the symbol, only a z-index to pull it to the foreground.

Force different screens to test all layouts. 

# Help requested | Aide requise

This needs to be tested on android.

Other home screens without cards have been tested. This should not break them. 

Initially we were working on new designs to avoid having to overlap shapes. So we now have two options to decide from!

### DIFF
--- a/src/components/RoundedBox.tsx
+++ b/src/components/RoundedBox.tsx
@@ -13,7 +13,6 @@ export const RoundedBox = ({children, isFirstBox}: RoundedBoxProp) => {
       style={isFirstBox ? [styles.roundedBox, styles.firstBox] : styles.roundedBox}
       marginBottom="m"
       alignSelf="stretch"
-      paddingTop="l"
     >
       <Box paddingHorizontal="m" paddingVertical="m">
         {children}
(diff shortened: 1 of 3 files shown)


amazingphilippe:
@smcmurtry Just an update on this. It seems to produce consistent results both on ios and android!

Capture d’écran, le 2021-03-09 à 09 31 51

timarney:
@amazingphilippe do we want to look at moving forward with this or do this work in a fresh PR?

amazingphilippe:
I think both approaches can work. This draft was a proof of concept to see how z-index could work on both platforms. 

However, if we want to do a fresh PR, we could tackle padding of the rounded box as clear scope, which could include the content of this draft + fixing other padding inconsistencies. Notably with bottom padding or spacing between paragraphs

timarney:
Closing - incorporated into https://github.com/cds-snc/covid-alert-app/pull/1497

Thanks @amazingphilippe this is a great update to bring iOS + Android inline.

PR #1429, opened 2021-02-18
New bottom menu design

smcmurtry:
# Summary | Résumé

In this PR I am killing off our custom `BottomSheet` component and making it a regular page that appears on top of the homescreen. This simplifies a lot of things. The design of the closed menu has changed to indicate that the menu is no longer "swipeable" and the contents of the menu have had a design refresh as well.

This PR also makes it possible to set the "QR Enabled" feature flag in the test menu rather than the .env file. I can pull that part out into a separate PR if desired, although it is somewhat related since it makes it easy to see how the menu changes depending on this setting.


src/components/Icon.tsx
```diff
@@ -62,11 +66,15 @@ const ICONS = {
...
   'canada-logo': CanadaLogo,
   'hand-reminder-red': HandReminderRed,
-  'qr-code': QRCode,
   'scan-qr-code-white-arrow': ScanQRCodeWhiteArrow,
   'camera-permission': Camera,
   'green-circle-check': GreenCircleCheck,
   'red-circle-exclamation': RedCircleExclamation,
+  'qr-code-icon': QRCodeIcon,
+  'hamburger-menu': HamburgerMenu,
+  close: Close,
```
timarney, line 75: Missing quotes here?

src/screens/menu/MenuScreen.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+          <Box marginBottom="m">
+            <ConditionalMenuPanels />
+          </Box>
+
+          <Box marginBottom="m" testID="InfoShareViewID">
+            <InfoShareItem
+              testID="getCodeButton"
+              onPress={onGetCode}
+              text={i18n.translate('Info.GetCode')}
```
timarney, line 80: Minor - but what if we moved the `text attr` to the top of all the  `InfoShareItem` components to make it a little easier to see what the item actually is.

smcmurtry: good idea, I will do that

src/components/Icon.tsx
smcmurtry: My linter told me not to. They are just keys in a javascript object so they don't need them, unless they are [not a valid js identifier](https://stackoverflow.com/questions/4348478/what-is-the-difference-between-object-keys-with-quotes-and-without-quotes), which is the case if there is a dash in the name.

timarney (approved):

PR #1435, opened 2021-02-22
Patch/adjust metric upload interval

jeberhardt:
# Summary | Résumé

This will adjust how often the Metrics will be uploaded to the server. Instead of relying on an interval of minutes to be passed, it will upload the metrics as long as it hasn't already been uploaded for that day.


craigzour (approved):

smcmurtry (approved):
Looks good.

PR #1438, opened 2021-02-22
Chore/screen size fix

omartehsin1:
# Summary | Résumé

Made some styling changes to various screens to accommodate smaller screens.


src/screens/qr/components/BaseQRCodeScreen.tsx
```diff
@@ -17,15 +17,17 @@ export const BaseQRCodeScreen = ({children, showBackButton, showCloseButton}: Ba
...
-            navText={i18n.translate('DataUpload.Close')}
-            showBackButton={showBackButton}
-            showCloseButton={showCloseButton}
-            useWhiteText={false}
-          />
-        </Box>
-        <View style={styles.flex}>{children}</View>
+        <ScrollView>
+          <Box marginBottom="m">
+            <Toolbar
```
smcmurtry, line 22: I think the Toolbar should be outside of the scrollview so the back and close buttons are always present at the top of the screen, similar to how the BaseDataSharingView works.

smcmurtry (changes requested):
I'm getting overlap between the Back button and the device status bar on my Moto G4: 
Screen Shot 2021-02-24 at 11 44 04 AM

Also, to match figma the background on this screen should have some transparency. I think the reason for that is that the QR code can be scanned even if it is outside of the frame, so seeing the QR code behind the black masked off area lets the user know that the phone can see it.

smcmurtry (approved):

PR #1443, opened 2021-02-25
Package bumps

timarney:
Updated semver + yarn


smcmurtry (approved):
Installed and tested on Android, seems to work fine.

PR #1452, opened 2021-03-02
Feature/exposure check using server time

craigzour:
# Summary | Résumé

- Added solution to force an exposure check if the local time on the device can't be trusted


craigzour:
@timarney Let's discuss about this PR tomorrow :-)

timarney:
Closing we'll revisit as needed.

PR #1454, opened 2021-03-03
Remove unused styles and colours from default.ts

smcmurtry:
# Summary | Résumé

There are a lot of styles and colours we are no longer using in default.ts. This PR removes them.


timarney:
@aklef can you test with with the visual diff script?

aklef:
> @aklef can you test with with the visual diff script?

This will be a wonderful first test of the script! Working on it!

aklef:
Here's one view of the output from the visual-diff script! Let me know if this is helpful in any way, or if you'd rather see the raw HTML somehow... Github doesn't allow me to attach HTML to an issue

[Visual Diff - visual-diff.ios vs test-pr1454.ios.pdf](https://github.com/cds-snc/covid-alert-app/files/6111331/Visual.Diff.-.visual-diff.ios.vs.test-pr1454.ios.pdf)

timarney:
Looks good - need to bring this up to date with latest master.

timarney (approved):

PR #1455, opened 2021-03-04
Publish active-user metric and send daily metrics as first step durin…

craigzour:
- Publish active-user event and send daily metrics at the very beginning of the background task

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationServiceProvider.tsx
+++ b/src/services/ExposureNotificationService/ExposureNotificationServiceProvider.tsx
@@ -58,6 +58,7 @@ export const ExposureNotificationServiceProvider = ({
 
   useEffect(() => {
     backgroundScheduler.registerPeriodicTask(async () => {
+      await FilteredMetricsService.sharedInstance().addEvent({type: EventTypeMetric.ActiveUser});
       await exposureNotificationService.updateExposureStatusInBackground();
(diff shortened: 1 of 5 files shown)


src/services/ExposureNotificationService/ExposureNotificationServiceProvider.tsx
```diff
@@ -58,6 +58,7 @@ export const ExposureNotificationServiceProvider = ({
 
   useEffect(() => {
     backgroundScheduler.registerPeriodicTask(async () => {
+      await FilteredMetricsService.sharedInstance().addEvent({type: EventTypeMetric.ActiveUser});
       await exposureNotificationService.updateExposureStatusInBackground();
```
timarney, line 62: Looks like we could move the add event + send inside the updateExposureStatusInBackground function (save some duplication)

craigzour: There is no more sendDailyMetrics here. I put it back at the end of `updateExposureStatusInBackground `. Also the active user event is not publish in `updateExposureStatusInBackground ` anymore.

craigzour: And now the active user is set up to be published right away

timarney (approved):

PR #1456, opened 2021-03-04
Added manufacturer and android release version to global metrics payload

craigzour:
# Summary | Résumé

- Added manufacturer and android release version to global metrics payload


CalvinRodo (commented):
I am aware, but can't review because I don't know enough about the app. 

We can handle this on the server.

timarney (approved):

PR #1467, opened 2021-03-23
Update packages - React Native 0.64.0

timarney:
React Native + Gradle updates as well as other dev dependancies


src/navigation/DevPersistedNavigationContainer.tsx
```diff
@@ -63,8 +63,8 @@ function DevPersistedNavigationContainerImpl(
 
   return (
     <NavigationContainer
-      {...others}
       key={persistKey}
+      {...others}
```
jeberhardt, line 67: What is the purpose of this change?

timarney: Updated linter says key prop needs to come before spread

jeberhardt (approved):

PR #1468, opened 2021-03-23
Test chore/update build packages

craigzour:
# Summary | Résumé

Testing something


craigzour (commented):
According to https://github.com/Borales/actions-yarn#github-actions-for-yarn we should now be using https://github.com/actions/setup-node

PR #1474, opened 2021-03-29
Remove `LazyThreadSafetyMode.NONE` parameter when loading by lazy.

jeberhardt:
# Summary | Résumé

A small change in how lazy variables are loaded in Kotlin, removing the `LazyThreadSafetyMode.NONE` parameter, which will now use the default `LazyThreadSafetyMode. SYNCHRONIZED`.

I don't believe this will have any noticeable impact on the app, as there don't seem to be any use cases where these objects would be accessed by multiple threads, but this is the safer approach to handling these variables.

Running the app on Android, and all normal behaviour should be executed.

### DIFF
--- a/android/app/src/main/java/app/covidshield/module/ExposureCheckSchedulerModule.kt
+++ b/android/app/src/main/java/app/covidshield/module/ExposureCheckSchedulerModule.kt
@@ -23,7 +23,7 @@ class ExposureCheckSchedulerModule(private val context: ReactApplicationContext)
 
     override val coroutineContext: CoroutineContext get() = Dispatchers.Default
 
-    private val workManager: WorkManager by lazy(LazyThreadSafetyMode.NONE) {
+    private val workManager: WorkManager by lazy {
         WorkManager.getInstance(context.applicationContext)
     }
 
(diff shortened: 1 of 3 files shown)


craigzour:
Could be the right PR to also tackle this if you think it can be an issue @jeberhardt @timarney 

I noticed something in one of the Kotlin module that is exposed to React.
Both @ReactMethod functions are being provided a promise that is never either resolved or rejected. The React layer uses both of them using the await  keyword.

https://github.com/cds-snc/covid-alert-app/blob/master/android/app/src/main/java/app/covidshield/module/ExposureCheckSchedulerModule.kt#L36

jeberhardt:
Nice catch @craigzour, but let's put that fix into a separate PR. The issues are not related.

craigzour (approved):

timarney:
ref:

> As default it will use SynchronizedLazyImpl (which correspond to LazyThreadSafetyMode.SYNCHRONIZED mode), so if you use it in this way you are going to be thread safe.

https://proandroiddev.com/kotlin-faster-lazy-for-android-7328ec8d8d57

timarney (approved):

PR #1476, opened 2021-03-31
Fix/metrics

craigzour:
# Summary | Résumé

- Fixed Exposed metric not being published at the right place
- Fixed OtkEntered metric being published every time we upload new TEKs

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -854,9 +854,16 @@ export class ExposureNotificationService {
       });
       return this.finalize({}, lastCheckedPeriod);
     }
+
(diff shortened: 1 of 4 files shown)


timarney:
Looks good - I'll let @smcmurtry give the thumbs up or down

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -854,9 +854,16 @@ export class ExposureNotificationService {
       });
       return this.finalize({}, lastCheckedPeriod);
     }
+
+    this.filteredMetricsService.addEvent({
```
smcmurtry, line 858: I think this is a good place to put it

smcmurtry (approved):
Looks good @craigzour!

PR #1477, opened 2021-04-01
Fixed EN toggle metric publishing. It now only gets published when th…

craigzour:
# Summary | Résumé

- Fixed EN toggle metric publishing. It now only gets published when the user manually turn it on or off.


src/screens/home/views/ExposureNotificationsDisabledView.tsx
```diff
@@ -16,7 +16,7 @@ export const ExposureNotificationsDisabledView = () => {
   }, []);
 
   const startEn = useCallback(async () => {
-    await startExposureNotificationService();
+    await startExposureNotificationService(true);
```
smcmurtry, line 19: change to `false`

src/screens/home/views/ExposureNotificationsUnauthorizedView.tsx
```diff
@@ -16,7 +16,7 @@ export const ExposureNotificationsUnauthorizedView = () => {
   }, []);
 
   const startEn = useCallback(() => {
-    startExposureNotificationService();
+    startExposureNotificationService(true);
```
smcmurtry, line 19: change to `false`

src/screens/home/views/LocationOffView.tsx
```diff
@@ -16,7 +16,7 @@ export const LocationOffView = () => {
   }, []);
 
   const startEn = useCallback(() => {
-    startExposureNotificationService();
+    startExposureNotificationService(true);
```
smcmurtry, line 19: change to `false`

src/screens/menu/views/SystemStatusUnauthorized.tsx
```diff
@@ -9,7 +9,7 @@ export const SystemStatusUnauthorized = () => {
   const startExposureNotificationService = useStartExposureNotificationService();
   const onPress = async () => {
     if (Platform.OS === 'android') {
-      await startExposureNotificationService();
+      await startExposureNotificationService(true);
```
smcmurtry, line 12: change to `false`

src/screens/menu/views/SystemStatusOff.tsx
```diff
@@ -9,7 +9,7 @@ export const SystemStatusOff = () => {
   const startExposureNotificationService = useStartExposureNotificationService();
   const onPress = async () => {
     if (Platform.OS === 'android') {
-      await startExposureNotificationService();
+      await startExposureNotificationService(true);
```
smcmurtry, line 12: change to `false`

smcmurtry (changes requested):
Since this metric is to figure out how much people are using the on/off feature, it should just be sent from these 2 places:
- the on/off button in the menu
- the "Turn back on" button on the `ExposureNotificationsUserStoppedView`

Let's remove it from the "on" buttons on the other error screens, since those are not specific to the on/off feature.

src/screens/menu/views/TurnAppBackOn.tsx
```diff
@@ -8,7 +8,7 @@ export const TurnAppBackOn = () => {
   const startExposureNotificationService = useStartExposureNotificationService();
 
   const onStart = useCallback(async () => {
-    await startExposureNotificationService();
+    await startExposureNotificationService(true);
```
craigzour, line 11: @smcmurtry You did not leave any comment here. Should I also set it to `false` ?

smcmurtry: I just checked and it looks like the `TurnAppBackOn` component isn't used anywhere. So it doesn't matter - this whole file could be deleted.

feralbob: s/could/should

😀

craigzour: Haha! I was in the process of doing it :)

PR #1480, opened 2021-04-06
Downgrade react-native-gesture-handler

timarney:
# Summary | Résumé

Navigating the menu via TalkBack (Android) is broken when using the latest version of `react-native-gesture-handler`

Rolling back to 1.8.0 

Tested up to `1.9.0` release where things start to fail

Potentially caused by  
https://github.com/software-mansion/react-native-gesture-handler/pull/1172/files#diff-e614cdf726640239cd9b5e551cfc1c50eb1d6f078b8c6884bfa4fb321ed82e4dR111

Original package bump happened here https://github.com/cds-snc/covid-alert-app/pull/1467 to fix an issue on iOS given the React Native version updates.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -313,8 +313,8 @@ PODS:
     - React-Core
   - RNCPushNotificationIOS (1.7.3):
     - React-Core
-  - RNGestureHandler (1.10.3):
-    - React-Core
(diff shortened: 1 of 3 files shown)


smcmurtry (approved):
Tested locally on Android and this solves the Ripple issue with Talkback for me 🎉

PR #1484, opened 2021-04-08
Generate Kotlin locale strings from JSON

timarney:
# Summary | Résumé

Adds script to generate `Android` XML string translation files from existing JSON locale files

We need to access translated strings from React and Kotlin. The source for the translated string should be in a single spot i.e. en.json. The existing json format will not work within Kotlin.

Generate xml files from the JSON files with Android / Kotlin can read.

> [Trello](https://trello.com/c/2gWHLAqq/1213-figure-out-how-to-get-translation-strings-currently-in-react-native-over-to-kotlin)

### Testing

String "paths" for xml generation are added via the `stringPaths` variable in the script

i.e.
```javascript
const stringPaths = ['AppName', 'NoExposureDetected.RegionCovered.Title'];
```
Run
```bash
yarn generate-translations:android
```

*View the updated Android XML files*

Screen Shot 2021-04-08 at 8 23 25 AM

*View the log output*

**Change the language on your Phone**

2 logs have been added to the `MainApplication.java` onCreate

You can replace  the existing `getResources().getString(R.string.app_name)` with a new string i.e.   `getResources().getString(R.string.the_string_name_here)`

```
@Override
    public void onCreate() {
        ...
        Log.d("locale", Locale.getDefault().getLanguage());
        Log.d("locale", getResources().getString(R.string.app_name));
       ...
    }
```

Screen Shot 2021-04-08 at 8 27 55 AM

### DIFF
--- a/android/app/src/main/java/app/covidshield/MainApplication.java
+++ b/android/app/src/main/java/app/covidshield/MainApplication.java
@@ -17,12 +17,14 @@ import org.jetbrains.annotations.NotNull;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 
 import org.unimodules.adapters.react.ModuleRegistryAdapter;
 import org.unimodules.adapters.react.ReactModuleRegistryProvider;
 import org.unimodules.core.interfaces.SingletonModule;
 
 import app.covidshield.generated.BasePackageList;
+import android.util.Log;
 
 
 
@@ -65,6 +67,8 @@ public class MainApplication extends Application implements ReactApplication {
     @Override
     public void onCreate() {
         instance = this;
+        Log.d("locale", Locale.getDefault().getLanguage());
+        Log.d("locale", getResources().getString(R.string.app_name));
         super.onCreate();
         SoLoader.init(this, /* native exopackage */ false);
     }
--- a/android/app/src/main/java/app/covidshield/MainApplication.java
+++ b/android/app/src/main/java/app/covidshield/MainApplication.java
@@ -67,8 +67,6 @@ public class MainApplication extends Application implements ReactApplication {
     @Override
     public void onCreate() {
         instance = this;
-        Log.d("locale", Locale.getDefault().getLanguage());
-        Log.d("locale", getResources().getString(R.string.app_name));
         super.onCreate();
         SoLoader.init(this, /* native exopackage */ false);
     }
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/app/covidshield/MainApplication.java
```diff
@@ -65,6 +67,8 @@ public ReactNativeHost getReactNativeHost() {
     @Override
     public void onCreate() {
         instance = this;
+        Log.d("locale", Locale.getDefault().getLanguage());
```
jeberhardt, line 70: The locale will need to come from the user selected and saved language, not the system locale.

timarney: Calling that out of scope for this PR.

Sticking just to the XML generating scripts for this one.

Mentioned in the Trello card:
> Out of scope
Updating the app to set the default language in Kotlin when the user changes the device language within the app

timarney: For another PR we should be able to use `Locale.setDefault(new Locale("fr_CA"));`

craigzour: @timarney @jeberhardt Here is how we can use the generated files https://github.com/cds-snc/covid-alert-app/commit/1585adb3fa667f5016e04e62fbf685f9eeb910d8

@timarney I did not add the commit in this PR as it uses code that is not yet been properly tested yet and is also a WIP

timarney: I think we could merge this as is and do a follow up PR for the other commit.

craigzour: @timarney I do agree with you! The follow up PR should be based on the branch we are currently working on with @jeberhardt (https://github.com/cds-snc/covid-alert-app/tree/feature/android-exposure-notification-port) because it is where we will need to plug the language stuff.

Before reviewing the PR, would you mind squashing the commits? I feel like it is almost all about the same `translations-android.js` script

craigzour, line 70: Now that we have an example on how we can use your generated files (https://github.com/cds-snc/covid-alert-app/commit/1585adb3fa667f5016e04e62fbf685f9eeb910d8) I think we could remove this.

craigzour (approved):

timarney:
@jeberhardt looks like we'll need your stamp of approval - @craigzour reviewed and tested but isn't a " code owner" yet.

PR #1490, opened 2021-04-14
Add app state listener + state to ExposureNotificationService

timarney:
Update to allow "non active" i.e. app running in the background to skip timestamp and onboarded check.

### DIFF
--- a/src/services/ExposureNotificationService/ExposureNotificationService.ts
+++ b/src/services/ExposureNotificationService/ExposureNotificationService.ts
@@ -660,6 +660,12 @@ export class ExposureNotificationService {
     return this.finalize();
   }
 
+  public isAppInBackground = () => {
+    if (this.appState !== 'active') {
+      return true;
+    }
+  };
+
   public shouldPerformExposureCheck = async () => {
     log.debug({
       category: 'exposure-check',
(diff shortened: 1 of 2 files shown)


src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -653,6 +661,19 @@ export class ExposureNotificationService {
...
+    log.debug({
+      category: 'exposure-check',
+      message: 'shouldPerformExposureCheck',
+      payload: {
+        appState: this.appState,
+      },
+    });
+
+    // pass thru unless the app is open
+    if (this.appState !== 'active') {
```
dj2, line 673: Could we pull this to a helper with a name like `isAppInBackground()` to make it clearer what we're checking?

craigzour (approved):
LGTM!

src/services/ExposureNotificationService/ExposureNotificationService.ts
timarney: Good idea - we can make a separate PR later to replace other instances as we see fit. 

```javascript
const onAppStateChange = async (newState: AppStateStatus) => {
      if (newState === 'active') {
        await fetchData();
      }
    };
```

dj2 (approved):

PR #1491, opened 2021-04-14
Feature/android debug metrics updates

jeberhardt:
# Summary | Résumé

Improving the debug metrics collection.

-- removing unnecessary `shouldPerformExposureCheck()` calls, since that function will always return **true** when the app is in the background.
-- removing Debug Metric for Step 6.1, as it's not needed.
-- adding extra information in 2 of the debug metric calls.

Get the app into a state where a background check should perform. Note that the metrics are still uploaded and the exposure check performs properly. Be sure to test when the app is in the task switcher, and when it's been killed altogether.


craigzour (approved):
LGTM

PR #1492, opened 2021-04-14
Feature/once per utc debug metrics

craigzour:
# Summary | Résumé

> Added an API to allow for once per UTC day debug metrics publishing

### DIFF
--- a/android/app/src/main/java/app/covidshield/services/metrics/UniqueDailyDebugMetricsHelper.kt
+++ b/android/app/src/main/java/app/covidshield/services/metrics/UniqueDailyDebugMetricsHelper.kt
@@ -1,0 +1,31 @@
+package app.covidshield.services.metrics
+
+import android.content.Context
+import app.covidshield.utils.DateUtils
+import java.util.*
+
+object UniqueDailyDebugMetricsHelper {
+
+    private const val SharedPreferencesFileKey = "covid-shield-unique-daily-debug-metrics"
+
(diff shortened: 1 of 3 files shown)


android/app/src/main/java/app/covidshield/services/metrics/UniqueDailyDebugMetricsHelper.kt
```diff
@@ -0,0 +1,40 @@
...
+    fun markMetricAsPublished(metricIdentifier: String, context: Context) {
+        val sharedPreferences = context.getSharedPreferences(SharedPreferencesFileKey, Context.MODE_PRIVATE)
+
+        with (sharedPreferences.edit()) {
+            putLong(metricIdentifier, System.currentTimeMillis())
+            apply()
+        }
+    }
+
+    private fun isSameDay(date1: Date, date2: Date): Boolean {
```
jeberhardt, line 30: It would be nice if this function was in a class on it's own, or with some other date functions.

jeberhardt (approved):

android/app/src/main/java/app/covidshield/services/metrics/UniqueDailyDebugMetricsHelper.kt
craigzour: @jeberhardt I put it here because I knew you were going to have some kind of a Date class at some point with the Kotlin rewrite but since we do not know if it is going to happen any time soon then I am going to extract it.

PR #1493, opened 2021-04-15
Added new debug metrics 200.0, 200.1 and 200.2 which will be sent onc…

craigzour:
…e per UTC day

# Summary | Résumé

> Added new debug metrics 200.0, 200.1 and 200.2 which will be sent once per UTC day


craigzour:
Going through initial testing

timarney (approved):

PR #1496, opened 2021-04-15
Move `setForeground()` to top of code.

jeberhardt:
# Summary | Résumé

In reviewing the code, it was observed that the `setForeground()` should happen before executing the call to react-native. This was highlighted in this article: https://medium.com/androiddevelopers/use-workmanager-for-immediate-background-execution-a57db502603d

It is hoped to solve the following errors we are occassionally seeing:
> Calls to setForegroundAsync() must complete before a ListenableWorker signals completion of work by returning an instance of Result.

Also added a try/catch surrounding blocks of code within `ExposureCheckSchedulerModule` to capture debugMetric errors if any exceptions are thrown at that step.

Running the app as normal. All background checks should continue as normal.  Look for the new error codes (108 & 109) in the debug metrics.


android/app/src/main/java/app/covidshield/receiver/worker/ExposureCheckNotificationWorker.kt
```diff
@@ -85,13 +68,32 @@ class ExposureCheckNotificationWorker (private val context: Context, parameters:
...
             val foregroundInfo = ForegroundInfo(1, notification.build())
             setForeground(foregroundInfo)
+        } catch (exception: Exception) {
+            MetricsService.publishDebugMetric(107.0, context, exception.message ?: "Unknown")
+            return Result.failure()
+        }
+
+        val enIsEnabled = exposureNotificationClient.isEnabled.await()
+        val enStatus = exposureNotificationClient.status.await()
+        if (!enIsEnabled || enStatus.contains(ExposureNotificationStatus.INACTIVATED)) {
```
dj2, line 78: Do we want to leave this check at the top? If EN is off, we should bail as early as possible?

jeberhardt: I think the goal here is to have the `setForeground()` call as soon as possible. Reading the comment in the medium.com article I linked to above.

dj2: Right, having it above the try below makes sense. I guess my question is, will the notification do anything that the user would notice? If they've disabled EN for whatever reason, we shouldn't be causing foreground work? Which is why I wonder if this specific check should still be at the top.

craigzour: PR looks good but I am just wondering why do we have to run this background task as a foreground service?

craigzour (approved):
LGTM

android/app/src/main/java/app/covidshield/receiver/worker/ExposureCheckNotificationWorker.kt
jeberhardt: It's the recommendation from Google to run Exposure Checks as a foreground service.

PR #1499, opened 2021-04-16
Feature/more debug metrics adjustments

craigzour:
# Summary | Résumé

- Added try/catch in ExposureCheckHeadlessTask with new debug metrics step 109
- Added new debug metrics 200.3 which will only be sent once per UTC day

### DIFF
--- a/android/app/src/main/java/app/covidshield/receiver/worker/ExposureCheckHeadlessTask.java
+++ b/android/app/src/main/java/app/covidshield/receiver/worker/ExposureCheckHeadlessTask.java
@@ -99,8 +99,14 @@ public class ExposureCheckHeadlessTask implements HeadlessJsTaskEventListener {
     }
 
     private void invokeStartTask(ReactContext reactContext, final HeadlessJsTaskConfig taskConfig) {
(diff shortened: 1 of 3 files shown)


craigzour:
Going through initial testing

android/app/src/main/java/app/covidshield/receiver/worker/ExposureCheckHeadlessTask.java
```diff
@@ -99,9 +99,14 @@ public void run() {
...
-            MetricsService.publishDebugMetric(3.9, reactContext);
-            return;
+
+        try {
+            if (reactContext.getLifecycleState() == LifecycleState.RESUMED) {
+                MetricsService.publishDebugMetric(3.9, reactContext);
+                return;
+            }
+        } catch (Exception exception) {
+            MetricsService.publishDebugMetric(109, reactContext, exception.getMessage());
```
dj2, line 109: I think we should return here, something has gone wrong and I'm not sure if we can really continue?

jeberhardt (approved):

PR #1505, opened 2021-04-19
Chore/scanned location

omartehsin1:
# Summary | Résumé

Updated Check In successful screen and invalid qr screen
Fixed some a11y issues.


timarney:
Couple of minor spacing issues

timarney:
We should should see about adding "expanded" text readout for the address i.e. currently reads 7:21 Am on A P R 20 2021 - we can make this pattern for the other pages where the address exists as well.

timarney:
Needs some margin at the bottom of the button for smaller screens.

timarney (approved):

PR #1509, opened 2021-04-20
Converted debug metrics 200.0 and 200.3 to keeper metrics

craigzour:
# Summary | Résumé

- Converted debug metrics 200.0 and 200.3 to keeper metrics


craigzour:
Will have to define real metric name before merging to master

timarney (approved):

PR #1512, opened 2021-04-21
Chore/a11y fix

omartehsin1:
# Summary | Résumé

Mainly fixed accessibility settings:
- Issues with autofocus on headings
- Accessibility label to read out "April" instead of "A-P-R"
- Accessibility label on the delete icon to read out, "Delete (business/event name)"


timarney (approved):
Nice - this is a really solid improvement compared to the current.

PR #1532, opened 2021-04-28
Allow a checkIn to be removed by locationId.

jeberhardt:
# Summary | Résumé

Updating the code to allow for the deletion of all check-ins as well as individual check-ins.

- Run the app, and enable the QR Feature from the debug menu.
- Check-in a few times using whatever QR codes you'd like.
- From the "Menu", select "Your Visits".
- Delete a single visit by selecting the Trash Can icon beside the visit.
- Delete all visits by selecting the "Delete" button at the bottom.


timarney:
Replace with: https://github.com/cds-snc/covid-alert-app/pull/1541

PR #1540, opened 2021-04-29
Add guard for severity text

timarney:
Given an app in an outbreak state pressing negative test + confirming was throwing an error given a property that wasn't available.


smcmurtry (approved):
Good catch!

PR #1552, opened 2021-05-03
Initial recent exposure screens

timarney:


smcmurtry (approved):
Built + tested on iPhone SE, looks like the outbreak screens work as expected

omartehsin1 (approved):

PR #1554, opened 2021-05-03
Moving the ExposureHistory to its own navigator and folder

smcmurtry:
# Summary | Résumé

What I did:
- Moved the ExposureHistoryScreen out of qr/ into its own folder
- Split out some views from the ExposureHistoryScreen into separate files
- created a new navigation stack for the ExposureHistoryScreen that includes the RecentExposureView
- swapped out the toolbar for one that contains a close button


src/screens/exposureHistory/RecentExposureView.tsx
```diff
@@ -26,10 +26,10 @@ export const RecentExposureScreen = () => {
   const i18n = useI18n();
 
   const navigation = useNavigation();
-  const back = useCallback(() => navigation.goBack(), [navigation]);
+  const close = useCallback(() => navigation.navigate('Home'), [navigation]);
   return (
     <SafeAreaView style={styles.flex}>
-      <Toolbar title="" navIcon="icon-back-arrow" navText={i18n.translate('PlacesLog.Back')} onIconClicked={back} />
+      <Toolbar2 navText={i18n.translate('DataUpload.Close')} showBackButton onIconClicked={close} />
```
timarney, line 32: Let's rename this to make it more clear as to the difference.

omartehsin1:
[Figma](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=19347%3A72175) only has the back button, this one has both close and the back button.

smcmurtry:
> [Figma](https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=19347%3A72175) only has the back button, this one has both close and the back button.

Yes - I diverged from figma there to be more consistent with the rest of the app.

src/screens/exposureHistory/RecentExposureView.tsx
smcmurtry: Good idea.

timarney (approved):

amazingphilippe:
Yep! All good for the close button. Just an oversight from me. Should the list page only have a close button since they essentially have the same destination?

```
+------------+    +-------------+   +-------------+                
|            |    |       close |   |<back  close |                
|    Menu    |--->|             |-->|             |                
|            |    | list      > |   |notification |                
+------------+    +-------------+   +-------------+
```

smcmurtry:
@amazingphilippe Yes that is a good point! Great ASCII art by the way :)

PR #1569, opened 2021-05-04
Update design for RecentExposureScreen

smcmurtry:
# Summary | Résumé

I changed the RecentExposureScreen to look more like figma, using custom components that borrow parts of the exposure views that appear on the home screen. I added a "Delete exposure" button that doesn't do anything yet.

Figma: https://www.figma.com/file/0pLawavG1fnBTXSYn4pdKZ/Exposure-notification?node-id=19149%3A515

RecentExposureScreen for Proximity:
Screen Shot 2021-05-04 at 4 16 35 PM

RecentExposureScreen for Outbreak:
Screen Shot 2021-05-04 at 4 16 52 PM


amazingphilippe:
The figma version may still be a bit wonky in terms of spacing too, but in general I think the spacing between heading and paragraph is too consistent here. We should reduce spacing between elements that belong together a bit!

Consistent spacing:
Capture d’écran, le 2021-05-05 à 09 33 27

Grouped:
Capture d’écran, le 2021-05-05 à 09 39 28

amazingphilippe:
Or even like this depending how the delete button is perceived in this context:

Capture d’écran, le 2021-05-05 à 09 42 28

smcmurtry:
@amazingphilippe I updated the spacing and changed the button (see below). It's not exactly what you want - there is extra space underneath the headings. We are re-using those components from the homescreen so it would be a lot of re-factoring to change that spacing. 

Screen Shot 2021-05-05 at 8 42 20 AM

omartehsin1 (approved):
Nice, looks good

PR #1574, opened 2021-05-05
bug fix for force outbreak screen

omartehsin1:
# Summary | Résumé

There was an issue where forcing an Outbreak Exposure view from the test menu resulted in a white screen and app crash.

What I did:

Added a default case to the switch statement to default to one of the Severity levels
'Date of exposure' will just be current date for the force screen


src/screens/home/views/OutbreakExposedView.tsx
```diff
@@ -19,7 +19,10 @@ export const OutbreakExposedView = () => {
   const historyItem: OutbreakHistoryItem = currentOutbreakHistory[0];
 
   const severity = historyItem?.severity;
-  const exposureDate = formatExposedDate(new Date(historyItem?.checkInTimestamp), dateLocale);
+  const exposureDate =
+    historyItem?.checkInTimestamp === undefined
+      ? formatExposedDate(new Date(), dateLocale)
```
timarney, line 24: We should only force a date if we're in TEST_MODE 

@smcmurtry fallback scenario - don't show a date ?  What do you think?

omartehsin1: updated

timarney (approved):

PR #1577, opened 2021-05-06
Enabled pushToServerRightAway option for QR code related metrics

craigzour:
# Summary | Résumé

- Enabled pushToServerRightAway option for QR code related metrics


timarney (approved):
Perfect :)

PR #1548, opened 2021-04-30
Updates and Fixes for e2e

aklef:
# Summary | Résumé

This PR proposes quite a few fixes for e2e tests, better documentation about them, what they are, how to run them, common issues, some help for those, and descriptions of what the files do in general. We now also properly support detox testing the new outbreak exposures and, for that matter, any future app screens.

# Test instructions | Instructions pour tester les modifications

Read the new docs, try to run the tests! Feedback on both the docs and the tests are appreciated

PR #1578, opened 2021-05-06
Align yarn run-ios simulator config with Detox configs

aklef:
# Summary | Résumé

Having completed #1548, I figured we should probably standardize which iPhone simulators are used in the yarn configs.

We already use the config proposed by this PR in `.detoxrc.json` This PR aims to align the yarn configs with that.

# Reviewer checklist | Liste de vérification du réviseur

Run `yarn run-ios` and verify that it opens with an iPhone 8 ( written in simulator status bar)


timarney:
I like this idea

@jeberhardt thoughts?

PR #1582, opened 2021-05-06
Added a new array to storage with proximity exposure history

smcmurtry:
# Summary | Résumé

Purpose:
The new "Recent exposures" screen contains a list of all recent outbreak and proximity exposures. These history items need to be displayed even if the user has already reported a negative test and cleared their status. This works for outbreak exposures because we are already storing a persistent `outbreakHistory`. However, our existing proximity exposure history (`exposureHistory`) is just an array of timestamps that gets cleared when the user reports a negative test. We can't use this because the proximity exposure history would disappear from the recent exposures page after a negative test. So I will implement a new proximity exposure history that does not get cleared and is displayed on this page.

What I did:
- Implemented a new `displayExposureHistory` that uses a new type definition `ProximityExposureHistoryItem[]`. This is the proximity exposure history displayed on the ExposureHistoryScreen. Variable names could be better. One challenge is that the existing `exposureHistory` would be better called the `proximityExposureTimestamps` or something like that. But I didn't want to mess with the existing code.
- Changed the `outbreakId` property to `id` so that it has the same name as the new id on `ProximityExposureHistoryItem`. This simplifies the logic on the RecentExposureScreen
- implemented a function `ignoreProximityExposure` that marks a proximity exposure as ignored, similar to how the outbreak history `isIgnored` works
- hooked everything up to the ExposureHistoryScreen and the delete button on the RecentExposureScreen

Still to do:
- ~change the "delete all exposures" button on the ExposureHistoryScreen to ignore all items from the displayExposureHistory~ done
- ~add logic to change your exposure status if a current exposure is deleted/ignored (follow up PR)~ done
- ~add an `isExpired` property to the history, and logic to expire entries 14 days after exposure (follow up PR)~ done
- ~add some migration code so people with existing exposures will get those transferred into the new storage object (follow up PR)~ done
- thorough testing
 
Testing:
- check out this branch
- check out either `ExposureNotificationAdapter.ios.ts` or `ExposureNotificationAdapter.android.ts` from branch `testing/force-proximity-exposure`, this will generate fake proximity exposures
- force a reload from metro, force a proximity exposure check from the test menu
- go exposed
- change `daysSinceLastExposure` to be `7` in the fakeSummary in the `ExposureNotificationAdapter` file
- force metro to reload, force proximity exposure check again
- observe there are now 2 entries in "Your recent exposures"
- test out the "delete exposures" buttons and the "delete all exposures" button


src/screens/exposureHistory/ExposureHistoryScreen.tsx
```diff
@@ -49,17 +49,18 @@ const toOutbreakExposureHistoryData = ({
...
+  proximityExposureHistory: ProximityExposureHistoryItem[];
   i18n: I18n;
 }): CombinedExposureHistoryData[] => {
-  return proximityExposureTimestamps.map(timestamp => {
+  return proximityExposureHistory.map(item => {
     return {
       exposureType: ExposureType.Proximity,
       subtitle: i18n.translate('QRCode.ProximityExposure'),
-      timestamp,
+      timestamp: item.notificationTimestamp,
```
timarney, line 62: Can we use `notificationTimestamp` here?

src/services/ExposureNotificationService/ExposureNotificationService.ts
```diff
@@ -958,6 +989,77 @@ export class ExposureNotificationService {
...
   }
 
+  public async saveDisplayExposureHistory() {
+    await this.storageService.save(
+      StorageDirectory.ExposureNotificationServiceDisplayExposureHistoryKey,
+      JSON.stringify(this.displayExposureHistory.get()),
+    );
+  }
+
+  /** this function is only for use on the ExposureHistoryScreen - it only effects the display logic */
```
timarney, line 999: Maybe add - a why this needs to exist after the initial comment

src/screens/exposureHistory/ExposureHistoryScreen.tsx
smcmurtry: I did this and immediately found a bug! I had been using the exposure timestamp for outbreaks on the history screen.

src/services/ExposureNotificationService/ExposureNotificationService.ts
smcmurtry: done

PR #1596, opened 2021-05-10
Re-render cache by using timestamp

timarney:
Negative test button was re-rendering homescreen


smcmurtry (approved):
Looks like the issue is fixed!

PR #1597, opened 2021-05-11
Adjustment to prevent menu text wrapping

timarney:
@smcmurtry please give this a test.

### DIFF
--- a/src/screens/menu/components/MenuButton.tsx
+++ b/src/screens/menu/components/MenuButton.tsx
@@ -37,6 +37,5 @@ const styles = StyleSheet.create({
   },
   menuBoxWrap: {
     minWidth: 200,
-    flexWrap: 'nowrap',
   },
 });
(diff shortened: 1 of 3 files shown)


src/screens/menu/components/MenuButton.tsx
```diff
@@ -33,4 +35,8 @@ const styles = StyleSheet.create({
     justifyContent: 'space-between',
     alignItems: 'center',
   },
+  menuBoxWrap: {
+    minWidth: 200,
```
smcmurtry, line 39: Looks like this is the change that actually fixes the issue. I reverted all the other changes and introduced this one and that fixed it.

src/screens/menu/components/MenuButton.tsx
```diff
@@ -33,4 +35,8 @@ const styles = StyleSheet.create({
...
+    flexWrap: 'nowrap',
```
smcmurtry, line 40: I don't think this is doing anything - at least on Android.

smcmurtry (approved):
Fixed! This is on my MotoG4 with window width of 360:

Screen Shot 2021-05-11 at 8 56 54 AM

src/screens/menu/components/MenuButton.tsx
timarney: Removed the flexWrap: 'nowrap',

PR #1600, opened 2021-05-11
Add button margin

timarney:
image


omartehsin1 (approved):
Good catch, looks better in landscape now.

PR #1604, opened 2021-05-11
Bug fix to ensure overlapping outbreak events do not result in more than one exposure per checkin

smcmurtry:
# Summary | Résumé

When we get outbreak events from the server that overlap the same day for a single location, this can result in multiple exposures for the same check in. This is visible to the user in the recent exposure history. We should only count this as one exposure, and display the guidance for the most severe of the outbreak events the user matched with. If there were overlapping outbreak events with the same severity, it doesn't matter which one we choose to keep so I am arbitrarily keeping the one with the latest `endTime`.

### DIFF
--- a/src/services/OutbreakService/OutbreakService.ts
+++ b/src/services/OutbreakService/OutbreakService.ts
@@ -10,6 +10,7 @@ import {unzip} from 'react-native-zip-archive';
 import {readFile} from 'react-native-fs';
 import {covidshield} from 'services/BackendService/covidshield';
 import {EventTypeMetric, FilteredMetricsService} from 'services/MetricsService';
+import {getRandomString} from 'shared/logging/uuid';
 
 import {Observable} from '../../shared/Observable';
 import {
@@ -31,6 +32,7 @@ export const HOURS_PER_PERIOD = 24;
 export const EXPOSURE_NOTIFICATION_CYCLE = 14;
 
 export interface OutbreakEvent {
+  id: string;
   locationId: string;
   // ms
   startTime: number;
@@ -212,6 +214,7 @@ export class OutbreakService {
   convertOutbreakEvents = (outbreakEvents: covidshield.OutbreakEvent[]): OutbreakEvent[] => {
     return outbreakEvents.map(event => {
       return {
+        id: getRandomString(8),
         locationId: event.locationId,
         endTime: 1000 * Number(event.endTime?.seconds),
         startTime: 1000 * Number(event.startTime?.seconds),
--- a/src/services/OutbreakService/OutbreakService.ts
+++ b/src/services/OutbreakService/OutbreakService.ts
@@ -32,7 +32,9 @@ export const HOURS_PER_PERIOD = 24;
 export const EXPOSURE_NOTIFICATION_CYCLE = 14;
 
 export interface OutbreakEvent {
-  id: string;
+  // Don't use this for anything besides the dedup code.
+  // dedupeId will change each time we get new data from the server.
+  dedupeId: string;
   locationId: string;
   // ms
   startTime: number;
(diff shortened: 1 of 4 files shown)


src/shared/qr.ts
```diff
@@ -245,3 +254,28 @@ export const getNewOutbreakExposures = (
   const newOutbreakExposures = detectedExposures.filter(item => newIds.indexOf(item.outbreakId) > -1);
   return newOutbreakExposures;
 };
+
+/**
+ * Look at all outbreak/checkin matches and remove duplicate matches
```
omartehsin1, line 259: By duplicate, this means outbreaks that have the same time stamp right? If I scan the same outbreak within a minute of each other I should see two outbreaks in my recent exposure?

maryrae-qa:
Retest completed on 3 Androids phones:
1. Android MotoG6 V9 (Dev 1.2.9 (311)), I5N1ZGZ7
2. Android Pixel 2 V9 (Dev 1.2.9 (311)), DCJZ7NVH
3. Android S7 V8.0 v12.5.1 (Dev 1.2.9 (311)), H3AV169C
Exposures were not duplicated for Proximity or Outbreak in the 'Your recent exposures' list.
**Retest Passed**

src/services/OutbreakService/OutbreakService.ts
```diff
@@ -212,6 +214,7 @@ export class OutbreakService {
   convertOutbreakEvents = (outbreakEvents: covidshield.OutbreakEvent[]): OutbreakEvent[] => {
     return outbreakEvents.map(event => {
       return {
+        id: getRandomString(8),
```
smcmurtry, line 217: Will add a comment here: "don't use this for anything besides the dedup code - it will change each time we get new data from the server"

Maybe rename to be: `dedupeId`

src/shared/qr.ts
```diff
@@ -245,3 +254,28 @@ export const getNewOutbreakExposures = (
...
+ * so the same checkin never results in more than one exposure
+ */
+export const deduplicateMatches = (allMatches: MatchData[]) => {
+  const deduplicationDict: MatchDeduplicationDict = {};
+  for (const match of allMatches) {
+    const timestampStr = match.timestamp.toString();
+    if (
+      Object.keys(deduplicationDict).indexOf(timestampStr) === -1 ||
+      deduplicationDict[timestampStr].maxOutbreakEndTimestamp < match.outbreakEvent.endTime ||
+      deduplicationDict[timestampStr].maxSeverity < match.outbreakEvent.severity
```
smcmurtry, line 269: replace these logic statements with named functions to make the code more readable

src/shared/qr.spec.ts
```diff
@@ -343,5 +349,70 @@ describe('outbreakHistory functions', () => {
...
+        outbreakEvent: outbreakEvent1,
+      };
+      const match2: MatchData = {
+        timestamp: checkIn1.timestamp,
+        checkIn: checkIn1,
+        outbreakEvent: outbreakEvent2,
+      };
+      expect(deduplicateMatches([match1, match2])).toStrictEqual([match2]);
+    });
+
```
smcmurtry, line 401: add test for same severity, different end time

src/shared/qr.ts
smcmurtry: Yes 2 scans that fall in the same outbreak should result in 2 exposures

src/shared/qr.spec.ts
smcmurtry: I removed this behaviour because it was buggy - now if 2 overlapping scans have the same severity, it will just pick the first one it iterates over.

src/shared/qr.ts
```diff
@@ -245,3 +254,35 @@ export const getNewOutbreakExposures = (
...
+
+/**
+ * Look at all outbreak/checkin matches and remove duplicate matches
+ * so the same checkin never results in more than one exposure
+ */
+export const deduplicateMatches = (allMatches: MatchData[]) => {
+  const deduplicationDict: MatchDeduplicationDict = {};
+  for (const match of allMatches) {
+    const timestampStr = match.timestamp.toString();
+    if (isKeyInObject(timestampStr, deduplicationDict) || isSeverityHigher(match, deduplicationDict)) {
```
timarney, line 275: Nice - this is much easier to understand now

timarney (approved):

omartehsin1 (approved):

PR #1606, opened 2021-05-13
Add + cleanup logging

timarney:
Added QR code logging to help debug

Cleaned up some noisy logging.

### DIFF
--- a/src/shared/qr.ts
+++ b/src/shared/qr.ts
@@ -126,7 +126,6 @@ export const getMatchedOutbreakHistoryItems = (
   checkInHistory: CheckInData[],
   outbreakEvents: OutbreakEvent[],
 ): OutbreakHistoryItem[] => {
-  log.debug({message: 'fetching outbreak locations', payload: {outbreakEvents}});
   const outbreakIds = outbreakEvents.map(event => event.locationId);
 
   const checkInLocationMatches = checkInHistory.filter(checkIn => {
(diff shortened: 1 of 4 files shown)


src/shared/qr.ts
```diff
@@ -126,7 +126,6 @@ export const getMatchedOutbreakHistoryItems = (
   checkInHistory: CheckInData[],
   outbreakEvents: OutbreakEvent[],
 ): OutbreakHistoryItem[] => {
-  log.debug({message: 'fetching outbreak locations', payload: {outbreakEvents}});
   const outbreakIds = outbreakEvents.map(event => event.locationId);
```
craigzour, line 129: As discussed earlier you may want to add a log here to dump the content of `checkInHistory` and `checkInHistory`. This would be purely for debug purposes though

omartehsin1 (approved):

src/shared/qr.ts
timarney: I moved it to where the fetch actually happens

https://github.com/cds-snc/covid-alert-app/pull/1606/files#diff-84f7d8a18f318b367ed570ef59acedc98e2ae98c9d53a717bfb899782076d68eR181

craigzour: Oh yeah! Did not see it :)

craigzour (approved):

PR #1607, opened 2021-05-13
fixed issue with special characters

omartehsin1:
# Summary | Résumé

Screen Shot 2021-05-13 at 1 40 42 PM

Screen Shot 2021-05-13 at 1 40 47 PM

There was an issue with the way the base64 string was being decoded. It would not render special characters correctly.

Added a function that will convert it back to it's original string before data is parsed

### DIFF
--- a/src/screens/qr/utils.ts
+++ b/src/screens/qr/utils.ts
@@ -31,7 +31,6 @@ const parseData = (data: string) => {
 
 export const handleOpenURL = async ({url}: EventURL): Promise<CheckInData> => {
   const [scheme, , host] = url.split('/');
-
   if (scheme !== 'https:' || host !== QR_HOST) {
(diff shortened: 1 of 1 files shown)


src/screens/qr/utils.ts
```diff
@@ -50,11 +49,12 @@ export const handleOpenURL = async ({url}: EventURL): Promise<CheckInData> => {
       base64Str = base64.encodeFromByteArray(data);
     }
 
-    const _locationData = base64.decode(base64Str);
+    const _locationData = b64Decodeutf8(base64.decode(base64Str));
```
timarney, line 52: Minor 

```
b64Decodeutf8
```

Maybe 


```
b64DecodeUtf8
```

Slightly easier to read?

timarney (approved):

PR #1609, opened 2021-05-14
Alternate exposure history implementation

smcmurtry:
# Summary | Résumé

This branch is based on `feature/proximity-exposure-history` (https://github.com/cds-snc/covid-alert-app/pull/1582) and has an alternate implementation where deletions on the history page effect the home page, and deletions on the home page remove entries from the history page. I will leave this as a draft and we can resurrect it if we decide to go that route.


smcmurtry:
Closing since I believe this is no longer needed.

PR #1619, opened 2021-05-21
iOS run script for production

timarney:
Creates new run-ios command for production


jeberhardt (approved):

jeberhardt:
Will need to be integrated in with the Fastlane scripts.

PR #1627, opened 2021-05-26
Introduced OutbreakSignatureValidationModule in Kotlin

craigzour:
Android - validate Outbreak payload signature from server.


timarney:
Ref: https://github.com/cds-snc/covid-alert-app/pull/1614

timarney (approved):

PR #1647, opened 2021-05-31
Chore/no notification diagnosed state

omartehsin1:
# Summary | Résumé

Fix for the issue that caused notifications to appear when in diagnosed state. This fix will. now precent the app from checking for exposures while in diagnosed state.

### DIFF
--- a/src/services/OutbreakService/OutbreakService.ts
+++ b/src/services/OutbreakService/OutbreakService.ts
@@ -365,8 +365,9 @@ export class OutbreakService {
   };
 }
 
-export const isDiagnosed = (status: string) => {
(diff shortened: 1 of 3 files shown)


src/services/OutbreakService/OutbreakService.ts
```diff
@@ -363,3 +364,9 @@ export class OutbreakService {
     return periodsToFetch;
   };
 }
+
+export const isDiagnosed = (status: string) => {
+  if (status === ExposureStatusType.Diagnosed) {
+    return true;
+  }
```
timarney, line 371: For good measure lets return false vs no value here and add a Boolean return type to the definition

timarney (approved):

PR #1660, opened 2021-06-02
updated en and fr content

omartehsin1:
# Summary | Résumé

Simply had to replace "Stop" and "Slow" with "limit"


EliseKa (approved):
perfect!

timarney (approved):

Issue #1663, opened 2021-06-03
QR Code Scanner / Camera - crashing app when background task runs


timarney:
Note: this feature hasn't been released yet.

The following error happens when the background task runs.

```
E/Camera2ClientBase: setPreviewCallbackFlag: attempt to use a locked camera from a different process (old pid 0, new pid 12897)
2021-06-03 13:48:08.502 771-1778/? E/Camera2ClientBase: getParameters: attempt to use a locked camera from a different process (old pid 0, new pid 12897)
2021-06-03 13:48:08.511 12897-14668/? E/AndroidRuntime: FATAL EXCEPTION: AsyncTask
```

<img width="1420" alt="Screen Shot 2021-06-03 at 1 50 17 PM" src="https://user-images.githubusercontent.com/62242/120695591-908a4600-c479-11eb-84bd-afe2bfc80158.png">

PR #1664, opened 2021-06-03
QR Code Scanner / Camera - crashing app fix

timarney:
Adds code to auto navigate Scanner / Camera view to home screen when the app moves to the background

Closes: https://github.com/cds-snc/covid-alert-app/issues/1663

### DIFF
--- a/src/screens/qr/views/QRCodeScanner.tsx
+++ b/src/screens/qr/views/QRCodeScanner.tsx
@@ -1,5 +1,5 @@
-import React, {useCallback, useState} from 'react';
-import {StatusBar, StyleSheet} from 'react-native';
+import React, {useCallback, useState, useEffect} from 'react';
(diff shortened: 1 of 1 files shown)


src/screens/qr/views/QRCodeScanner.tsx
```diff
@@ -46,6 +46,21 @@ export const QRCodeScanner = () => {
...
+        // see: https://github.com/cds-snc/covid-alert-app/issues/1663
+        navigation.navigate('Home');
+      }
+    };
+
+    AppState.addEventListener('change', onAppStateChange);
+    return () => {
+      AppState.removeEventListener('change', onAppStateChange);
+    };
+  }, []);
```
omartehsin1, line 62: Missing dependency, otherwise looks great.

omartehsin1 (approved):

PR #1683, opened 2021-06-09
Auto delete visits after 28 days

timarney:
# Summary | Résumé

Adds code to auto remove checkin after the CHECKIN_NOTIFICATION_CYCLE (28 days) has past.

### DIFF
--- a/src/services/OutbreakService/OutbreakService.spec.ts
+++ b/src/services/OutbreakService/OutbreakService.spec.ts
@@ -330,4 +330,26 @@ describe('OutbreakService', () => {
     const outbreakHistory = service.outbreakHistory.get();
     expect(outbreakHistory).toHaveLength(1);
   });
+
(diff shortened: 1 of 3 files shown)


src/services/OutbreakService/OutbreakService.spec.ts
```diff
@@ -330,4 +330,21 @@ describe('OutbreakService', () => {
     const outbreakHistory = service.outbreakHistory.get();
     expect(outbreakHistory).toHaveLength(1);
   });
+
+  it('auto deletes checkin after CHECKIN_NOTIFICATION_CYCLE', async () => {
+    await service.addCheckIn(checkIns[4]);
+    await service.addCheckIn(checkIns[5]);
+
+    // day 1
+    MockDate.set('2021-01-01T10:00Z');
```
omartehsin1, line 339: I changed the date on this to 2021-02-08, and it is still passing. The time stamps on the ID is 2021-01-01, shouldn't it fail?

omartehsin1 (approved):

PR #1684, opened 2021-06-09
Chore/should perform outbreak check tests

omartehsin1:
# Summary | Résumé

Tests the shouldPerformOutbreakCheck.
 - The date is: 2021-02-01T12:00:00.000Z, first check is 30 Mins after so should equal false.
 - The second test is 3 hours later and should return false.
 - The second test is 6 hours later and should return true.

### DIFF
--- a/src/services/OutbreakService/OutbreakService.spec.ts
+++ b/src/services/OutbreakService/OutbreakService.spec.ts
@@ -330,4 +330,32 @@ describe('OutbreakService', () => {
     const outbreakHistory = service.outbreakHistory.get();
     expect(outbreakHistory).toHaveLength(1);
   });
+
+  it('perform outbreak check', async () => {
+    jest.spyOn(service, 'extractOutbreakEventsFromZipFiles').mockImplementation(async () => {
+      return service.convertOutbreakEvents([
+        {
+          locationId: checkIns[0].id,
+          startTime: {seconds: toSeconds(subtractHours(checkIns[0].timestamp, 2))},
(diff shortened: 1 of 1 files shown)


src/services/OutbreakService/OutbreakService.spec.ts
```diff
@@ -330,4 +330,32 @@ describe('OutbreakService', () => {
...
+    await service.addCheckIn(checkIns[0]);
+    await service.addCheckIn(checkIns[1]);
+
+    await service.checkForOutbreaks();
+    // Current Date: 2021-02-01T12:00:00.000Z
+    MockDate.set('2021-02-01T12:30Z');
+    // First check 30 minutes later
+    const performOutbreakCheckFalse = await service.shouldPerformOutbreaksCheck();
+    expect(performOutbreakCheckFalse).toStrictEqual(false);
+
```
timarney, line 355: Maybe add one more in-between check 3hrs or something

src/services/OutbreakService/OutbreakService.spec.ts
```diff
@@ -330,4 +330,32 @@ describe('OutbreakService', () => {
     const outbreakHistory = service.outbreakHistory.get();
     expect(outbreakHistory).toHaveLength(1);
   });
+
+  it('perform outbreak check', async () => {
```
timarney, line 334: Let's make the test name more descriptive
i.e. 
performs and outbreak check if past the min minutes threshold

timarney (approved):

PR #1689, opened 2021-06-10
Chore/one check multiple exposures

omartehsin1:
# Summary | Résumé

This test is based on a situation where a user scans multiple exposed locations before the next background check, we want to ensure all outbreak within that timeframe are added to the history.


timarney (approved):
Nice :)

PR #1696, opened 2021-06-11
Remove exposed-outbreak metric

timarney:
# Summary | Résumé

Removing the exposed-outbreak this will not be sent to the metrics aggregation.


craigzour (commented):
Should we remove the rest of the code related to the metric?

timarney:
otk-entered-v2 will still exist so I think we still need the Event type etc..

craigzour (approved):

PR #1699, opened 2021-06-14
Fix/blank region value

craigzour:
# Summary | Résumé

- Fixed for Region set to "" instead of "None" in both native layers


timarney:
What's the best way to test this?

craigzour:
@timarney I did test it myself but I guess we could just this build on several devices where we do not select any region during the onboarding. Then we would have to check on the server side if we are getting metrics with a region set to "" (which should not be the case anymore)

timarney:
Going to approve and we'll test in the next staging builds.

timarney (approved):

patheard:
Happy to check for test payloads if you let me know what I should be looking for in DynamoDB.

PR #1705, opened 2021-06-15
Reworked the way we make sure iOS metrics are being sent once per UTC…

craigzour:
# Summary | Résumé

- Trying a new solution to save the fact that we already published once per UTC day type of metrics in iOS (native layer)


craigzour:
Currently in testing phase

jeberhardt (approved):

PR #1706, opened 2021-06-15
Refactored periodSinceLastCheck

omartehsin1:
# Summary | Résumé

We had a duplicate function in Outbreak and ExposureNotifications. It has just been refactored for a cleanup.


src/shared/date-fns.ts
```diff
@@ -156,3 +156,20 @@ export const getUTCMidnight = (date: Date) => {
   const midnight = new Date(Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate()));
   return midnight.getTime();
 };
+
+export const periodsSinceLastExposureFetch = (_lastCheckedPeriod?: number): number[] => {
```
timarney, line 160: Nice - think we should add a "type" parameter here which would either be proximity and outbreak

That would in turn use the const 
EXPOSURE_NOTIFICATION_CYCLE or OUTBREAK_NOTIFICATION_CYCLE (if for whatever reason it could be different in the future)

I'm concerned that given it's in the function we won't remember to update it `exposureNotificationCycle = 14;`

Adding type also allows us to do different things when testing.

timarney:
@omartehsin1 

I made some small adjustments to see if we could get better Type checking.

This should now show an error if your not passing a value that in the `enum`

Screen Shot 2021-06-16 at 8 52 45 AM

timarney (approved):

PR #1710, opened 2021-06-16
Chore/date locale clean up

omartehsin1:
# Summary | Résumé

There were many cases where we were creating  `dateLocale`  like: `const dateLocale = i18n.locale === 'fr' ? 'fr-CA' : 'en-CA';`. This could be prone to errors.

This branch keeps it in one single file and it can be imported whenever needed.


timarney:
This is throwing errors

Screen Shot 2021-06-16 at 12 51 52 PM

timarney (approved):

PR #1717, opened 2021-06-21
added proximity and outbreak tests

omartehsin1:
# Summary | Résumé

Tests to ensure:

- if we have both a proximity and exposure, we should be able to see both items in our history
- After receiving a negative test from an outbreak exposure (and also having an proximity exposure), your homescreen should return to "All set", the reason being the user has already seen the instruction on how to take precautions and has gotten tested.
- Last test is to check to see that both tests have expired and been removed from the history, status will stay as monitoring.


timarney (approved):
👍

PR #1718, opened 2021-06-21
Change `failure` to `success`.

jeberhardt:
# Summary | Résumé

Android only.

There is an issue affecting some users, where the notification that appears when **Checking for Notifications** does not disappear. It has been difficult to troubleshoot and debug since we have been unable to replicate it. One clue we have found is a Crash Report on the Google Play Console that happens when `Result.failure()` is called. Since we do not actively handle the `failure` case (opting to just let exposure checking happen again on the regular schedule), we believe it is safe and reasonable to replace the `failure()` call with a `success()` call.

Trello card:
https://trello.com/c/eJEE7ItN/1306-checking-for-exposures-notification-persists-on-android

Normal regression testing would apply in this case.


jeberhardt:
['Checking for exposures' notification persists on Android](https://trello.com/c/eJEE7ItN/1306-checking-for-exposures-notification-persists-on-android)

timarney (approved):

PR #1720, opened 2021-06-23
Fixed issue where outbreak exposure history is wiped upon launch

craigzour:
# Summary | Résumé

- Fixed issue where outbreak exposure history is wiped upon launch


omartehsin1 (approved):
Nice, tested and all seems to be working!

PR #1721, opened 2021-06-23
Android: Fixing persistent notifications.

jeberhardt:
# Summary | Résumé

An attempt to fix the "persistent notification" bug in Android. Adding a `withTimeout` block as suggested by Marcel from Google. Documentation here: https://developers.google.com/android/exposure-notifications/implementation-guide#doWork

Also double checking for the existence of the react-native context, and firing a headless event even though the code returned if it didn't exist. This is more in line with the previous way this was handled in the code.

Trello card: https://trello.com/c/eJEE7ItN/1306-checking-for-exposures-notification-persists-on-android

Normal regression testing to ensure exposure checks are still happening, and that the visual notification is still appearing as usual. Since we have been unable to replicate this issue, it will be difficult to know if this fixes the issue, but we should be testing to ensure that nothing new is broken.


jeberhardt:
['Checking for exposures' notification persists on Android](https://trello.com/c/eJEE7ItN/1306-checking-for-exposures-notification-persists-on-android)

timarney (approved):

PR #1746, opened 2022-02-09
feat: s3-backup remediation

CalvinRodo:
# Description

Adding the workflow that will back up this repo to a centralized s3 bucket. 

As part of our disaster recovery work we backup all of our github repos to an s3 bucket in the SRE Tools repo.

This workflows uses organizational centralized credentials to authenticate so there is no need for secret management on your side.

## Important information

This workflow is not mandatory, however if does make it easy to backup your project, and helps us improve our disaster recovery posture. If you do not want to use it will be expected that you achieve this goal another way.

This PR is not-critical and so it can wait.

This PR was generated by an automated tool. 

If you have a question about this PR you can use the PR review process or send a message to the Internal SRE Team.

### DIFF
--- a/.github/workflows/s3-backup.yml
+++ b/.github/workflows/s3-backup.yml
@@ -1,0 +1,30 @@
+name: S3 backup
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 6 * * *"
+
+env:
+  BACKUP_BRANCH: main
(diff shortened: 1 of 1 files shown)


patheard (approved):

CalvinRodo:
Bug in workflow :(

